PR #164, opened 2020-03-16
Custom validations expanded

antoniogarrote:
- Adding support for logical composition of custom validations
- Adding new types of custom validation constraints
- Removing JS custom validations
- Removing TopQuadrant SHACL library and using pure Java Jena one


looseale:
Merged from another branch

PR #291, opened 2020-06-11
OAS 3.0 - Emit status code for referenced responses in endpoint

AgustinBettati:


tomsfernandez (approved):
Remove the useless and big description

PR #308, opened 2020-06-19
Fixing errors in JS interface

antoniogarrote:
Matching change for https://github.com/mulesoft/amf-core/pull/65

- Fixing some other JS interfaces with the wrong namespace or old names.
- Updating the TS typings.


nschejtman:
No need for namespace changes

PR #530, opened 2020-07-14
Re-work documentation

jstoiko:
Summary of changes:

- reworded description (used the one on a.ml's homepage)
- clarified "vision"
- listed all natively supported "formats" and "models" under one section
- added a note on "Guaranteed output"
- listed all documentation-related links under one section
- renamed "RAML dialect" -> "AML dialect"
- moved contribution-related docs (e.g. tests) to CONTRIBUTING.md
- replaced (private) Jenkins build link/badge by Travis CI public one


README.md
```diff
@@ -2,66 +2,56 @@
...
-## General scope
-The library supports many of the required uses cases:
-- Parse a 0.8 / 1.0 RAML, 2.0 / 3.0 OAS and JSON-LD AMF model.
-- AMF API design model creation.
-- Model edition.
-- Export to any of the mentioned standards.
+- Support for multiple formats with a unified programming interface and model
+- Support for both syntactic and semantic model layers
+- Support for validation at both syntactic and semantic model layers
+- Produce formal specifications of any (specification) language
```
antoniogarrote: Provide formal specification for the supported specification languages?

README.md
```diff
@@ -2,66 +2,56 @@
...
-sbt coverage test coverageReport
-```
-### Generate artifacts directly from cloned repository
+The **only** guaranteed output of AMF is the JSON-LD "AMF model". Any other output such as any output provided natively by the models listed under the section above may change at any time. This means that while the semantic representation of those outputs may remain unchanged, the syntactical expression such as the order in which the outputted metadata is expressed and any other syntax-related constructs may change from one version of AMF to another. If this is an issue for your use-case, you may consider using a custom resolution/generation pipeline.
 
-```sh
-sbt package
-```
-This will generate *jvm* JARs in each of the module's targets.
+## Documentation
```
antoniogarrote, line 42: Should we link the link to the new amf-metadata repository with the models? https://github.com/aml-org/amf-metadata/

antoniogarrote (dismissed):
Left some comments for consideration

tomsfernandez:
Hi @jstoiko. Please cherry-pick your changes and rebase with develop so we can merge this.

README.md
jstoiko: specification ... specification - I think I'll remove this sentence, we're more or less saying this (or implying it) already.

jstoiko:
@tomsfernandez: I tweaked the Travis CI section. I think you should consider re-writing this section, it sounds a bit rough:

> ### Tests aren’t optional
> Any bugfix that doesn’t include a test proving the existence of the bug being fixed, may be suspect. Same for new features that can’t prove they actually work.

tomsfernandez:
Hi @jstoiko, how would you rephrase that?

jstoiko:
I would remove this section altogether. It is already very clear that tests are expected (i.e. `Contributions must comply with a minimum of 80% coverage rate`). In my experience, contributions that don't come with tests initially may be coming from less experienced contributors (regardless of what's written in the contributing guidelines...) and will require a bit more hand holding.

tomsfernandez:
Hi @jstoiko. The issue I have with removing that section is that a contributor may modify code that because of AMF's test suite has over 80% coverage. We should be upfront and say that tests are expected. I think that coding something _testable_ and proving so is a big part of a new contribution and should be addressed since the beginning of the contribution process.

It is true that new contributors may find it difficult to include tests. We could add something about that aspect but I wouldn't erase that every contribution needs test that assert the modified behaviour.

jstoiko:
My initial comment wasn't about having or not such mention but about the tone -- i.e. "Tests aren’t optional", "may be suspect" -- which may not sound very "inviting" (to contribute), to say the least. So maybe rephrasing this to something more "inviting" such as "please include tests with any PR submission" would do.

tomsfernandez:
Yep, it is better that way. If you want you can commit that modification or I can do it after merging this PR.

Please cherry-pick your changes si that I can merge this PR.

tomsfernandez:
Cherry-picked commits + extra changes in https://github.com/aml-org/amf/pull/654.

Closing this PR.

PR #571, opened 2020-07-31
APIMF-2308: avoid duplicate id in examples of external json schema

AgustinBettati:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasRefParser.scala
```diff
@@ -158,6 +158,7 @@ class OasRefParser(map: YMap,
             tmpShape.unresolved(text, e, "warning").withSupportsRecursion(true)
             Some(tmpShape)
           case Some(jsonSchemaShape) =>
+            //TODO: parsed json schema should be registered and linked, currently remote schemas are parsead repeatedly
```
tomsfernandez: Aren't schemas registered in the parseRemoteSchema function?

tomsfernandez (dismissed):

tomsfernandez (approved):

PR #584, opened 2020-08-11
ALS-1217: Definition keys annotation on documents

agusayerza:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/Raml08DocumentParser.scala
```diff
@@ -105,6 +115,7 @@ case class Raml08DocumentParser(root: Root)(implicit override val ctx: RamlWebAp
 
   private def parseSchemaDeclarations(map: YMap, parent: String): Unit = {
     map.key("schemas").foreach { e =>
+      ctx.addDeclarationKey(DeclarationKey(AnyShapeModel, e))
```
tomsfernandez, line 118: why not ShapeModel instead of AnyShapeModel?

tomsfernandez (dismissed):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/Raml08DocumentParser.scala
agusayerza: Based on the fact that the Raml08TypeParser returns an Option[AnyShape]. Do you think ShapeModel would be more appropriate?

tomsfernandez (approved):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/annotations/Annotations.scala
```diff
@@ -145,4 +147,29 @@ case class Inferred() extends Annotation
 
 case class CollectionFormatFromItems() extends Annotation
 
-case class ExternalJsonSchemaShape(original: YMapEntry) extends Annotation
\ No newline at end of file
+case class ExternalJsonSchemaShape(original: YMapEntry) extends Annotation
+
+case class DeclarationKeys(keys: List[DeclarationKey]) extends Annotation
+
+case class DeclarationKey(model: DomainElementModel,
```
pope1838, line 154: En lugar de model podriamos usar el uri del term aca? de esa manera serviria para declaraciones de dialect instances no?

PR #590, opened 2020-08-13
Include autotagging for support branches

AgustinBettati:


tomsfernandez:
We can use this opportunity to fix error codes in tagging script. If for any reason version tagging does not work it will not throw an error in Jenkins as the scripts last line is "echo" which will almost always return exit code 0.

We should use a few ifs or use the && to chain the statements and provide the correct error code for the operation

tomsfernandez (approved):

PR #591, opened 2020-08-13
APIMF-1964: validate 'consumes' property in a parameter of type 'file'

arielmirra:
add tests to validate 'consumes' property in a parameter of type 'file'


AgustinBettati:
These cases are being validated in OasJsonModelUniquePlatformReportTest with the following tests:
- Parameter of type file with no consumes property defined
- Parameters of type file defined in path with invalid consumes property
- Parameters of type file defined in operation with invalid consumes property

Issue #364, opened 2020-07-01
Creating and using a custom Oas20Plugin


postatum:
* **I'm submitting a ...**
  - [ ] question

Let's say I have a lib (written in Scala) which uses AMF and I want to create a custom `Oas20Plugin` to change the way OAS 2.0 emission works.

I imagine it as:
1. Copy and edit contents of `amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/OasPlugin.scala`
2. Register new plugin with `amf.Core.loadPlugin()`.

My questions are:
I. Do the steps I described make sense?
II. If I follow these steps and use `Oas20Renderer` would my plugin take precedence over the default `Oas20Plugin`? Or should I also create a custom `Oas20Renderer`?

postatum:
> Depends on what you want to achieve. What is your concrete use case?

I want to partially undo what was done by this commit  https://github.com/aml-org/amf/commit/f4a82948d76b7415e08130bd49f94612d6b39785. In particular, I want to create OAS 2-3 emitters/contexts/etc which would allow me to output OAS 2-3 the old way - without using "definitions" node.

nschejtman:
Ok, is this blocking you to achieve what you need to achieve? Why do you need inlined definitions rather extracted definitions?

postatum:
Hi @nschejtman. Thanks for answering this issue.

> Ok, is this blocking you to achieve what you need to achieve?

It does.

> Why do you need inlined definitions rather extracted definitions?

We have a lib [webapi-parser](https://github.com/raml-org/webapi-parser) which wraps AMF to provide users simplified interface. One of its features is the [resolution](https://raml-org.github.io/webapi-parser/resolving.html). In a nutshell it's a process of making specs super-flat - flattening types inheritance, moving types definitions to the places they are referenced from, etc.

With the addition of extracted definitions to OAS 2-3, "resolution" doesn't produce "flat" specs any more. I know it's a fault of emission logic and not resolution, but from users perspective it's the other way around.

You can find more details in the old AMF repo issue 596.

nschejtman:
Great, I understand. Since you have already identified the commit I suggest that rather than extending the Oas20Plugin you parametrize this change in emission

postatum:
Hi @nschejtman.

Thanks a lot for helping me with this!

Few questions:
[...]

PR #597, opened 2020-08-20
Add ability to disable OAS compacted emission

postatum:
This PR addresses issues discussed here https://github.com/aml-org/amf/issues/364. In particular it makes possible to disable OAS 2.0, OAS 3.0 "compacted" emission added by this commit https://github.com/aml-org/amf/commit/f4a82948d76b7415e08130bd49f94612d6b39785.

Compacted emission can be disabled with `RenderOptions().withoutCompactedEmission`.

amf-core part of necessary changes is here https://github.com/aml-org/amf-core/pull/98

I suggest you squish the PR as it contains a lot of unnecessary commits I did when trying to figure out how things work.


postatum:
The build fails because `amf-core` PR was not merged/released yet:

```
[error] /home/travis/build/aml-org/amf/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/OasPlugin.scala:172:17: value isWithCompactedEmission is not a member of amf.core.emitter.RenderOptions

[error]     if (options.isWithCompactedEmission) {
```

AgustinBettati:
Hello @postatum, thank you for opening the PR. 
I have a few comments to add regarding the changes you have made. You will see that OasSpecEmitterContext has a Boolean variable to disable compactEmission. Using this variable directly will significantly reduce the amount of code that has to be changed, there is no need to create the InlinedContext which forces you to to handle specific cases in emission. You will only have to modify RenderOptions, both OasPlugins in the specContext method, and include tests.
It would be great if your could update both of your branches (in amf and amf-core) with these adjustments and squash them into a single commit.

postatum:
> Hello @postatum, thank you for opening the PR.
> I have a few comments to add regarding the changes you have made. You will see that OasSpecEmitterContext has a Boolean variable to disable compactEmission. Using this variable directly will significantly reduce the amount of code that has to be changed, there is no need to create the InlinedContext which forces you to to handle specific cases in emission. You will only have to modify RenderOptions, both OasPlugins in the specContext method, and include tests.
> It would be great if your could update both of your branches (in amf and amf-core) with these adjustments and squash them into a single commit.

Hi @AgustinBettati!
Thanks for the feedback. I reworked the PR the way you've suggested and created a different PR with squished commits: https://github.com/aml-org/amf/pull/668

PR #606, opened 2020-08-26
ALS-1232 add lexical information to shapes from raml type expressions…

pope1838:
… arrays


BlockLucas (approved):

looseale:
This will be merge from release/4.3.0 (already cherry-picked there). The commit has JSON-LDs that are modified by new changes in develop, so looks like it will generate more merge conflicts if we have it in develop too.

PR #612, opened 2020-08-27
APIMF-1975: validate 'schema' property in a parameter in 'body'

arielmirra:
APIMF-1975: validate 'schema' property in a parameter in 'body'

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
@@ -430,10 +430,10 @@ case class Oas2ParameterParser(entryOrNode: YMapEntryLike,
     // Force to re-adopt with the new mediatype if exists
     if (payload.mediaType.nonEmpty) validateEntryName(payload)
 
-    map.key(
-      "schema",
-      entry => {
-        OasTypeParser(entry, shape => setName(shape).asInstanceOf[Shape].adopted(payload.id))(toOas(ctx)) // i don't need to set param need in here. Its necesary only for form data, because of the properties
+    map.key("schema") match {
+      case Some(entry) =>
+        // i don't need to set param need in here. Its necessary only for form data, because of the properties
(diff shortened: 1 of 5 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
```diff
@@ -433,19 +433,23 @@ case class Oas2ParameterParser(entryOrNode: YMapEntryLike,
...
               schema.annotations += ParameterBindingInBodyLexicalInfo(range)
             }
-            schema
+            val result = schema
+            result
           }
       }
     )
-
+    if (payload.schema == null) {
```
tomsfernandez: Can't this be done with an AMFRawValidation?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
```diff
@@ -433,19 +433,23 @@ case class Oas2ParameterParser(entryOrNode: YMapEntryLike,
...
+        OasTypeParser(entry, shape => setName(shape).asInstanceOf[Shape].adopted(payload.id))(toOas(ctx))
           .parse()
           .map { schema =>
             checkNotFileInBody(schema)
             payload.set(PayloadModel.Schema, tracking(schema, payload.id), Annotations(entry))
             bindingRange.foreach { range =>
               schema.annotations += ParameterBindingInBodyLexicalInfo(range)
             }
-            schema
+            val result = schema
```
tomsfernandez: Return result only

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
arielmirra: we tried but it's not possible

tomsfernandez (approved):

PR #629, opened 2020-09-08
APIMF-1880: RAML - validate target value in 'allowedTargets' field

arielmirra:
This issue had to be resolved in parsing because the value of the 'allowedTargets' has to be a valid string, and if it's later parsed, the value of that string can't be retreived.
For example:
String "Method" when parsed becomes a full Operation model, thus we can't obtain the string value in the API spec.

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
@@ -569,6 +569,46 @@ abstract class RamlSpecParser(implicit ctx: RamlWebApiContext) extends WebApiBas
   }
 
   case class AnnotationTypesParser(ast: YPart, annotationName: String, map: YMap, adopt: CustomDomainProperty => Unit) {
+
+    def checkValidTarget(entry: YMapEntry, nodeId: String): Unit = {
+      val targets = entry.value.value match {
+        case value: YScalar =>
+          Seq(value.text)
+        case values: YSequence =>
+          values.nodes.map(node => node.asScalar.get.text)
+      }
+
+      val validTargets: Set[String] = Set(
+        "API",
+        "DocumentationItem",
+        "Resource",
+        "Method",
+        "Response",
+        "RequestBody",
+        "ResponseBody",
+        "TypeDeclaration",
+        "Example",
+        "ResourceType",
+        "Trait",
+        "SecurityScheme",
+        "SecuritySchemeSettings",
+        "AnnotationType",
+        "Library",
+        "Overlay",
+        "Extension"
+      )
+
+      targets.foreach(target => {
+        if (!validTargets.contains(target))
+          ctx.eh.warning(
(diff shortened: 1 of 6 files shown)


amf-webapi/shared/src/main/scala/amf/validations/ParserSideValidations.scala
```diff
@@ -679,6 +684,7 @@ object ParserSideValidations extends Validations {
     ReadOnlyPropertyMarkedRequired.id    -> all(WARNING),
     MissingRequiredFieldForGrantType.id  -> all(WARNING),
     OasInvalidParameterSchema.id         -> all(WARNING), // should be violation
+    InvalidAllowedTargets.id             -> all(WARNING), // should be violation
```
tomsfernandez: Add as TODO to find it faster.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
```diff
@@ -569,6 +569,46 @@ abstract class RamlSpecParser(implicit ctx: RamlWebApiContext) extends WebApiBas
...
+
+    def checkValidTarget(entry: YMapEntry): Unit = {
+      val targets = entry.value.value match {
+        case value: YScalar =>
+          Seq(value.text)
+        case values: YSequence =>
+          values.nodes.map(node => node.asScalar.get.text)
+      }
+
+      val validTargets = Seq(
```
tomsfernandez: should be a Set to make lookup faster. In a BitSet, lookup has constant time.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
```diff
@@ -569,6 +569,46 @@ abstract class RamlSpecParser(implicit ctx: RamlWebApiContext) extends WebApiBas
   }
 
   case class AnnotationTypesParser(ast: YPart, annotationName: String, map: YMap, adopt: CustomDomainProperty => Unit) {
+
+    def checkValidTarget(entry: YMapEntry): Unit = {
+      val targets = entry.value.value match {
```
tomsfernandez, line 574: entry.element is also available

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
```diff
@@ -569,6 +569,46 @@ abstract class RamlSpecParser(implicit ctx: RamlWebApiContext) extends WebApiBas
...
+        "Library",
+        "Overlay",
+        "Extension"
+      )
+
+      targets.foreach(target => {
+        if (!validTargets.contains(target))
+          ctx.eh.warning(
+            InvalidAllowedTargets,
+            "",
```
tomsfernandez: the node id is missing

tomsfernandez (approved):

PR #633, opened 2020-09-10
APIMF-1962: OAS2 - warn when example field is present in a parameter

arielmirra:
should become a closed shape violation in the next major release

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
@@ -291,8 +291,8 @@ case class Oas2ParameterParser(entryOrNode: YMapEntryLike,
     }
   }
 
-  def parse(): OasParameter =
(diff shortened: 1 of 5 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ParameterParsers.scala
```diff
@@ -309,6 +309,23 @@ case class Oas2ParameterParser(entryOrNode: YMapEntryLike,
...
+  }
+
+  def checkExampleField(p: OasParameter): Unit =
+    map.key("example") match {
+      case Some(_) =>
+        /* TODO: Should remove 'example' from syntax and delete this method and parser-side validation
+                 as it will become a model validation and violation in the next major */
+        ctx.eh.warning(
+          invalidExampleFieldWarning,
+          p.domainElement.id,
```
tomsfernandez, line 323: should be parameter, not "p.domainElement.id" if the p is a Parameter

tomsfernandez (approved):

PR #641, opened 2020-09-17
Als fixes/lexical info for non patch

pope1838:


nschejtman (dismissed):

amf-webapi/shared/src/main/scala/amf/plugins/domain/webapi/models/templates/Trait.scala
```diff
@@ -42,7 +42,8 @@ class Trait(override val fields: Fields, override val annotations: Annotations)
               keepEditingInfo = false,
               errorHandler = errorHandler
             ))
-          .orNull // TODO should we return null or an empty operation?
+          .getOrElse(Operation()) // TODO should we return null or an empty operation?
+      // Yes, we should (an error handler is provided for situations
```
tomsfernandez: Remove TODO if already solved

tomsfernandez (dismissed):

PR #702, opened 2020-10-26
APIMF-2536: change CLI ErrorHandler

arielmirra:
APIMF-2536: change CLI ErrorHandler
IMPORTANT NOTE: The CLI will now throw violations that ommited before (breaking change?)

### DIFF
--- a/amf-client/shared/src/main/scala/amf/client/commands/CommandHelper.scala
+++ b/amf-client/shared/src/main/scala/amf/client/commands/CommandHelper.scala
@@ -50,8 +50,7 @@ trait CommandHelper {
       Option(effectiveMediaType(config.inputMediaType, config.inputFormat)),
       config.inputFormat,
       Context(platform),
-      cache = Cache(),
-      errorHandler = UnhandledParserErrorHandler
+      cache = Cache()
     )
     val vendor = effectiveVendor(config.inputFormat)
     if (config.resolve)
@@ -69,8 +68,7 @@ trait CommandHelper {
         Option(effectiveMediaType(config.inputMediaType, config.inputFormat)),
         config.inputFormat,
         Context(platform),
-        cache = Cache(),
-        errorHandler = UnhandledParserErrorHandler
+        cache = Cache()
       )
       parsed map { parsed =>
         RuntimeResolver.resolve(vendor, parsed, ResolutionPipeline.DEFAULT_PIPELINE, UnhandledErrorHandler)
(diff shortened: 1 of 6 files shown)


amf-client/shared/src/main/scala/amf/client/commands/CommandHelper.scala
```diff
@@ -50,8 +50,7 @@ trait CommandHelper {
       Option(effectiveMediaType(config.inputMediaType, config.inputFormat)),
       config.inputFormat,
       Context(platform),
-      cache = Cache(),
-      errorHandler = UnhandledParserErrorHandler
```
tomsfernandez, line 54: This change is strange, do we know the real reason why this is failing? Beside the symptom.

arielmirra: Yes, the short version is that the `UnhandledParserErrorHandler ` only throws the validations, and omits adding them to the `StaticErrorCollector`. 

Inside `UnhandledParserErrorHandler` you can see that it just throws an exception and nothing more:
```scala
override def handlerAmfResult(result: AMFValidationResult): Boolean = {
    throw new Exception(
      s"  Message: ${result.message}\n  Target: ${result.targetNode}\nProperty: ${result.targetProperty.getOrElse("")}\n  Position: ${result.position}\n at location: ${result.location}")
  }
```

tomsfernandez: Right but why doesn't the CLI's execution stop on an Exception?

arielmirra: that's something to analyze. I think it keeps validating and then returns all warnings/violations. But none are returned because none are stored.

amf-client/shared/src/test/scala/amf/client/commands/CommandLineTests.scala
```diff
@@ -113,6 +113,36 @@ class CommandLineTests extends AsyncFunSuite with PlatformSecrets {
...
+    val proc   = new TestProc()
+
+    ValidateCommand(platform).run(
+      cfg.get.copy(
+        stdout = stdout,
+        stderr = stderr,
+        proc = proc
+      )) map { _ =>
+      assert(stderr.acc == "")
+      assert(stdout.acc != "")
```
tomsfernandez, line 141: Shouldn't this test be more specific?

tomsfernandez (approved):

PR #705, opened 2020-10-27
Als 1251/payload annotations

agusayerza:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasContentParser.scala
```diff
@@ -35,7 +35,7 @@ case class OasContentParser(entry: YMapEntry, producer: Option[String] => Payloa
     val mediaTypeText: String = mediaTypeNode.text().toString
```
tomsfernandez: [Out of scope] a getMediaType method could be here that gets the key and returns a string.

tomsfernandez: [Out of scope] "buildPayload" name could describe better what it does.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasResponseParser.scala
```diff
@@ -100,10 +100,12 @@ case class OasResponseParser(map: YMap, adopted: Response => Unit)(implicit ctx:
...
         // OAS 3.0.0
         if (ctx.syntax == Oas3Syntax) {
-          map.key(
-            "content",
-            entry => payloads ++= OasContentsParser(entry, res.withPayload).parse()
-          )
+          map
+            .key("content")
+            .map(entry => {
+              res.set(ResponseModel.Payloads, AmfArray(payloads, Annotations(entry)))
```
tomsfernandez: This won't work as it will be overriden in line 123. I would patch it in the if in line 122 if you want to finish it fast. You can also refactor this shit class if you like also to run the change nicely.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasResponseParser.scala
```diff
@@ -100,10 +100,12 @@ case class OasResponseParser(map: YMap, adopted: Response => Unit)(implicit ctx:
 
         // OAS 3.0.0
         if (ctx.syntax == Oas3Syntax) {
```
tomsfernandez, line 103: [Out of scope] - Paraphrasing "this class is shit". The OasResponseParser needs a good refactor.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasResponseParser.scala
```diff
@@ -100,10 +100,12 @@ case class OasResponseParser(map: YMap, adopted: Response => Unit)(implicit ctx:
...
-          )
+          map
+            .key("content")
+            .map(entry => {
+              res.set(ResponseModel.Payloads, AmfArray(payloads, Annotations(entry)))
+              payloads ++= OasContentsParser(entry, res.withPayload).parse()
+            })
 
           map.key(
             "links",
```
tomsfernandez, line 112: [Out of scope] - The commented code on line 129 should be deleted. Intellij won't let me comment the right line

tomsfernandez (changes requested):

tomsfernandez:
Please squash both commits as both code and json-ld changes should be together as the golden changes are a direct consequence of the code change.

agusayerza:
Was thinking of doing a squash and merge, I have squashed them locally now

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasResponseParser.scala
agusayerza: Removed

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasResponseParser.scala
agusayerza: Could be overwritten but not necessarily. The idea behind the patch is to handle the case where we have the key "content" but the value is empty, resulting in an empty payload(*) list. The spec would still be valid, an AMF provided no real way to know if the key was present or not. 

* Except for the case where the RAML response payload had been provided.

tomsfernandez (approved):

PR #715, opened 2020-11-02
Apimf 2566/raml10

pope1838:


amf-client/shared/src/test/scala/amf/compiler/RamlTypeExpressionAnnotationTest.scala
```diff
@@ -45,14 +45,14 @@ class RamlTypeExpressionAnnotationTest extends AsyncFlatSpec with Matchers with
   }
 
   // todo: Should work when APIMF-2323 is solved
```
tomsfernandez: Erase TODO

amf-client/shared/src/test/scala/amf/compiler/RamlTypeExpressionAnnotationTest.scala
```diff
@@ -45,14 +45,14 @@ class RamlTypeExpressionAnnotationTest extends AsyncFlatSpec with Matchers with
...
-  ignore should "annotate range for union with three elements" in {
+  it should "annotate range for union with three elements" in {
     test(tripleUnionA)
     test(tripleUnionB)
     test(tripleUnionC)
   }
 
   // todo: Should work when APIMF-2323 is solved
```
tomsfernandez: Erase TODO

amf-client/shared/src/test/scala/amf/resolution/ProductionResolutionTest.scala
```diff
@@ -39,14 +39,14 @@ class ProductionResolutionTest extends RamlResolutionTest {
...
 
-  multiGoldenTest("Resolves googleapis.compredictionv1.2swagger.raml to jsonld",
-                  "googleapis.compredictionv1.2swagger.raml.resolved.%s") { config =>
-    cycle("googleapis.compredictionv1.2swagger.raml",
-          config.golden,
-          RamlYamlHint,
-          renderOptions = Some(config.renderOptions),
-          target = Amf)
-  }
+//  multiGoldenTest("Resolves googleapis.compredictionv1.2swagger.raml to jsonld",
```
tomsfernandez: why is this test commented out?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/annotations/Annotations.scala
```diff
@@ -143,8 +143,6 @@ case class OperationTraitEntry(range: Range) extends Annotation
 // save original text link?
 case class ReferencedElement(parsedUrl: String, referenced: DomainElement) extends Annotation
 
-case class Inferred() extends Annotation
```
tomsfernandez, line 146: Where did this annotation go?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/common/AbstractVariables.scala
```diff
@@ -1,21 +1,24 @@
...
+import amf.core.parser.{Annotations, ScalarNode}
 import amf.core.resolution.VariableReplacer.VariableRegex
-import org.yaml.model.YScalar
+import org.yaml.model.YNode
 
 import scala.collection.mutable
 
 case class AbstractVariables() {
-  val variables: mutable.Set[String] = mutable.Set()
+  private val variables: mutable.Set[(String, Annotations)] = mutable.Set()
```
tomsfernandez: We can't have a Map here because variable names may be repeated right? The same variable may be used in different places also?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/common/DataNodeParser.scala
```diff
@@ -54,17 +54,16 @@ class DataNodeParser private (node: YNode,
...
-      val key = ast.key.as[YScalar].text
-      parameters.parseVariables(key)
+      parameters.parseVariables(ast.key)
       val value               = ast.value
       val propertyAnnotations = Annotations(ast)
 
       val propertyNode =
         new DataNodeParser(value, refsCounter, parameters, Some(node.id), idCounter).parse().forceAdopted(node.id)
-      node.addProperty(key, propertyNode, propertyAnnotations)
+      node.addProperty(ast.key.as[YScalar].text, propertyNode, propertyAnnotations)
```
tomsfernandez, line 66: [Out of Scope] - 
```scala
ast.key.as[YScalar].text
```
could be in a separate method. I'm probably being picky about this.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/AbstractDeclarationParsers.scala
```diff
@@ -80,35 +72,40 @@ case class AbstractDeclarationParser(declaration: AbstractDeclaration, parent: S
       case Left(link) => parseReferenced(declaration, link, entryValue, map.annotations).adopted(parent)
       case Right(value) =>
         val variables = AbstractVariables()
-        val parentUri =
-          if (parent.contains("#")) s"$parent/$key"
-          else s"$parent#/$key"
+        named(declaration)
```
tomsfernandez, line 75: should be wrapped in same function to convey that those methods go together. This is for the despistado that might change the order of the methods and mess up id adoption.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/Raml10TypeParser.scala
```diff
@@ -317,7 +317,7 @@ case class Raml08UnionTypeParser(shape: UnionShape, types: Seq[YNode], ast: YPar
       .filter(_.isDefined)
       .map(_.get)
 
-    shape.setArray(UnionShapeModel.AnyOf, unionNodes, Annotations(ast))
```
tomsfernandez, line 320: [Out of Scope] - If we are going to stop using `setArray` because of Annotations we should remove it from the codebase.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/Raml10TypeParser.scala
```diff
@@ -1580,30 +1591,36 @@ sealed abstract class RamlTypeParser(entryOrNode: YMapEntryLike,
...
               )
 
             case _ =>
           }
 
-          property.set(
-            PropertyShapeModel.Path,
-            AmfScalar((Namespace.Data + entry.key.as[YScalar].text.urlComponentEncoded).iri(), Annotations(entry.key)))
+          property.set(PropertyShapeModel.Path,
+                       AmfScalar((Namespace.Data + entry.key.as[YScalar].text.urlComponentEncoded).iri()),
```
tomsfernandez, line 1609: [Out of Scope] - `entry.key.as[YScalar].text.urlComponentEncoded).iri()` extract to function

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/emitters/raml/RamlScalarShapeEmitter.scala
```diff
@@ -39,7 +39,7 @@ case class RamlScalarShapeEmitter(scalar: ScalarShape, ordering: SpecOrdering, r
     val typeEmitterOption = if (scalar.inherits.isEmpty) {
       fs.entry(ScalarShapeModel.DataType)
         .flatMap(f =>
-          if (!f.value.annotations.contains(classOf[Inferred])) {
+          if (!f.value.value.annotations.contains(classOf[Inferred])) {
```
tomsfernandez: `f.element.element`

tomsfernandez: [OOF] - Could be extracted to `isInferred` method also

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
```diff
@@ -215,20 +219,23 @@ abstract class RamlDocumentParser(root: Root)(implicit val ctx: RamlWebApiContex
...
     map.key("securedBy", (WebApiModel.Security in api using RequirementParser).allowingSingleValue)
     map.key(
       "documentation",
       entry => {
-        api.set(WebApiModel.Documentations,
-                AmfArray(UserDocumentationsParser(entry.value.as[Seq[YNode]], ctx.declarations, api.id).parse()),
-                Annotations(entry))
+        api.set(
+          WebApiModel.Documentations,
+          AmfArray(UserDocumentationsParser(entry.value.as[Seq[YNode]], ctx.declarations, api.id).parse(),
```
tomsfernandez, line 235: OOF - `entry.value.as[Seq[YNode]]` shouldn't be as[YSequence]?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtensionDomainElementMerge.scala
```diff
@@ -104,7 +104,7 @@ class ExtensionDomainElementMerge(restrictions: MergingRestrictions,
       case amfObject: AmfObject => (amfObject.id, amfObject.annotations)
       case array: AmfArray =>
         val ann =
-          if (value.annotations.nonEmpty) value.annotations
+          if (value.annotations.nonEmpty && !value.annotations.contains(classOf[VirtualNode])) value.annotations
```
tomsfernandez, line 107: Extract to function?

tomsfernandez (commented):
Left some comments and cosmetic changes.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/annotations/Annotations.scala
pope1838: I had to moved to amf-core

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/common/AbstractVariables.scala
pope1838: Exactly! If you have only one value, the others source maps will be lost.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/Raml10TypeParser.scala
pope1838: Not necessary. Some client methods can set just values in a memory model. In those cases, setArray maybe could be useful.
We should not use setArray in parsing clases. Another approach could be force the setArray method to build the AMfArray with virtual element annotation, but it could be weird for the client building model use case.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/raml/RamlDocumentParser.scala
pope1838: YSequence is an YPart that contains a Seq[YNode]. If you are going to use only the values, you can do as[Seq[YNode]].
In fact, that code did not change.
This is the discussion of YNode vs YNode.value in annotations.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/Raml10TypeParser.scala
pope1838: MMM not sure. Keeping aside that the code was already there, I think is a specific use-case of the YScalar conversion.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtensionDomainElementMerge.scala
pope1838: The virtual Node is an interface for any kind of virtual annotations, so I'm not sure if is good idea to generalize it.
Good options could be that the field is inferred or an AmfElement is a virtual element, but the general annotation as virtual sounds confused.

amf-client/shared/src/test/resources/org/raml/parser/fragments/library/output.txt
```diff
@@ -7,6 +7,7 @@ types:
     additionalProperties: true
```
tomsfernandez, line 7: why isn't this a yaml/raml file?

looseale (approved):

amf-client/shared/src/test/resources/org/raml/parser/fragments/library/output.txt
pope1838: To be honest, I don't remember. About 3 years ago, when we integrated the old raml parser we've chosen that kind of file. All the suit works that way, maybe we can talk about the suit, but is not in the scope of this change.

tomsfernandez (approved):

PR #734, opened 2020-11-17
APIMF-2679: parse message header into new header object field with no…

pope1838:
…de shape instead of build a virtual parameter

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/client/model/domain/Message.scala
+++ b/amf-webapi/shared/src/main/scala/amf/client/model/domain/Message.scala
@@ -30,6 +30,7 @@ class Message(override private[amf] val _internal: InternalMessage)
   def title: StrField                     = _internal.title
   def summary: StrField                   = _internal.summary
   def bindings: MessageBindings           = _internal.bindings
+  def headerSchema: NodeShape             = _internal.headerSchema
 
   /** Set name property of this Response. */
   override def withName(name: String): this.type = {
--- a/amf-webapi/shared/src/main/scala/amf/client/model/domain/Message.scala
+++ b/amf-webapi/shared/src/main/scala/amf/client/model/domain/Message.scala
@@ -88,6 +88,11 @@ class Message(override private[amf] val _internal: InternalMessage)
     this
   }
 
+  def withHeaderSchema(schema: NodeShape): this.type = {
+    _internal.withHeaderSchema(schema)
+    this
+  }
+
   def withPayload(mediaType: ClientOption[String]): Payload = _internal.withPayload(mediaType.toScala)
 
   def withPayload(): Payload = _internal.withPayload()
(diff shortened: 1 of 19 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/async/parser/AsyncMessageParser.scala
```diff
@@ -149,9 +150,18 @@ abstract class AsyncMessagePopulator()(implicit ctx: AsyncWebApiContext) extends
...
-        parseHeaderSchema(entry, message.id) foreach { param =>
-          message.set(MessageModel.Headers, AmfArray(Seq(param), Annotations(entry.value)), Annotations(entry))
+      entry => {
+        AsyncApiTypeParser(entry, shape => shape.withName("schema").adopted(message.id), JSONSchemaDraft7SchemaVersion)
+          .parse()
+          .foreach {
+            case n: NodeShape =>
+              message.set(MessageModel.HeaderSchema, n, Annotations(entry))
+            case _ =>
+              ctx.eh.violation(ParserSideValidations.HeaderMustBeObject,
```
tomsfernandez, line 162: [OOF] - A good improvement is to stop parsing the shape once we know its a nodeShape as we don't assign it the instance if it isn't a NodeShape. Probably not significant but still a processing saving

amf-webapi/shared/src/main/scala/amf/plugins/domain/webapi/models/Message.scala
```diff
@@ -28,6 +28,7 @@ class Message(override val fields: Fields, override val annotations: Annotations
   def bindings: MessageBindings    = fields.field(Bindings)
   def headerExamples: Seq[Example] = fields.field(HeaderExamples)
   def headers: Seq[Parameter]      = fields.field(Headers)
+  def headerObject: NodeShape      = fields.field(HeaderSchema)
```
tomsfernandez: headerObject should also be renamed to HeaderSchema

tomsfernandez (changes requested):
Please also update client side interface.

[OOF] - Update typings jeje

amf-webapi/shared/src/main/scala/amf/client/model/domain/Message.scala
```diff
@@ -30,6 +30,7 @@ class Message(override private[amf] val _internal: InternalMessage)
   def title: StrField                     = _internal.title
   def summary: StrField                   = _internal.summary
   def bindings: MessageBindings           = _internal.bindings
+  def headerSchema: NodeShape             = _internal.headerSchema
```
AgustinBettati, line 33: missing withHeaderSchema method in client interface!

AgustinBettati (changes requested):

tomsfernandez (approved):

AgustinBettati (approved):

PR #749, opened 2020-11-19
APIMF-2599: improve source maps annotations for oas 2.0 parsers

pope1838:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/SecurityRequirementsParsers.scala
```diff
@@ -119,7 +129,7 @@ case class OasLikeSecurityRequirementParser(node: YNode,
   private def getScopes(schemeEntry: YMapEntry) =
     schemeEntry.value
       .as[Seq[YNode]]
-      .map(n => Scope(n).set(ScopeModel.Name, AmfScalar(n.as[String]), Annotations(n)))
+      .map(n => Scope(n).set(ScopeModel.Name, ScalarNode(n).text(), Annotations.inferred()))
```
AgustinBettati, line 132: Lexical information which was previously present is lost

AgustinBettati (changes requested):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasLikeOperationParser.scala
```diff
@@ -114,12 +108,12 @@ abstract class OasOperationParser(entry: YMapEntry, adopt: Operation => Operatio
           .filter(y => !isOasAnnotation(y.key.as[YScalar].text))
           .foreach {
             entry =>
-              val node = ScalarNode(entry.key).text()
+              val node = ScalarNode(entry.key)
               responses += OasResponseParser(
                 entry.value.as[YMap], { r =>
-                  r.set(ResponseModel.Name, node, Annotations(entry.key))
+                  r.withName(node)
```
AgustinBettati, line 115: looks like lexical info is lost here by looking at oas30api.expanded.jsonld change

AgustinBettati (changes requested):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasLikeOperationParser.scala
pope1838: The method with name sets the field as inferred (because key "name" is not present, only the value).
On the other hand, Value.value annotations (name value) should be present, right?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/SecurityRequirementsParsers.scala
pope1838: Same here, lexical info for value.value should be present right?
We lost invalid value annotations (the one of the field) that now is maked as inferred

AgustinBettati (approved):

amf-client/shared/src/test/scala/amf/maker/DocumentMakerTest.scala
```diff
@@ -42,7 +43,7 @@ class DocumentMakerTest extends WebApiMakerTest {
 
   private def assertFixture(expected: Document, file: String, hint: Hint): Future[Assertion] = {
     Validation(platform).flatMap { v =>
-      AMFCompiler(basePath + file, platform, hint, eh = UnhandledParserErrorHandler)
+      AMFCompiler(basePath + file, platform, hint, eh = IgnoreErrorHandler)
```
tomsfernandez, line 46: Why start ignoring errors in this test?

amf-client/shared/src/test/scala/amf/maker/ReferencesMakerTest.scala
```diff
@@ -82,15 +82,15 @@ class ReferencesMakerTest extends AsyncFunSuite with CompilerTestBuilder with Am
...
       val personLink = person.link("fragments/" + fragmentFile).asInstanceOf[NodeShape].withName("person")
       if (recursive) personLink.withSupportsRecursion(true)
-
+      val api = WebApi()
+        .withId("amf-client/shared/src/test/resources/references/" + file + "#/web-api")
+        .withName("API")
+        .withVersion("1.0")
+      if (spec == Oas20 || spec == Oas) api.withEndPoints(Nil)
       Document()
         .withId("/Users/hernan.najles/mulesoft/amf/amf-client/shared/src/test/resources/references/" + file)
```
tomsfernandez: Change id for a non-specific one

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasTypeParser.scala
```diff
@@ -702,10 +708,11 @@ case class OasTypeParser(entryOrNode: YMapEntryLike,
...
         item  <- OasTypeParser(entry, items => items.adopted(shape.id + "/items"), version).parse()
       } yield {
+        shape.set(ArrayShapeModel.Items, item, Annotations(entry))
         item match {
-          case array: ArrayShape   => shape.withItems(array).toMatrixShape
-          case matrix: MatrixShape => shape.withItems(matrix).toMatrixShape
-          case other: AnyShape     => shape.withItems(other)
+          case _: ArrayShape  => shape.toMatrixShape
+          case _: MatrixShape => shape.toMatrixShape
+          case _: AnyShape    => shape
```
tomsfernandez: Unnecessary to match the last AnyShape

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasTypeParser.scala
```diff
@@ -970,11 +977,17 @@ case class OasTypeParser(entryOrNode: YMapEntryLike,
 
       val property = producer(name)
         .add(Annotations(entry))
-        .set(PropertyShapeModel.MinCount, AmfScalar(if (required) 1 else 0), requiredAnnotations += ExplicitField())
+        .set(
+          PropertyShapeModel.MinCount,
+          AmfScalar(if (required) 1 else 0),
+          Annotations
+            .synthesized() += ExplicitField()) // this field is not explicit. In fact, explicit field annotation should disappear? Keeping annotation for oas to raml conversion tests.
```
tomsfernandez: add TODO so we can identify this later on.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasTypeParser.scala
```diff
@@ -970,11 +977,17 @@ case class OasTypeParser(entryOrNode: YMapEntryLike,
...
 
       property.set(
         PropertyShapeModel.Path,
-        AmfScalar((Namespace.Data + entry.key.as[YScalar].text.urlComponentEncoded).iri(), Annotations(entry.key)))
+        AmfScalar((Namespace.Data + entry.key.as[YScalar].text.urlComponentEncoded).iri(), Annotations(entry.key)),
```
tomsfernandez: Extract `entry.key.as[YScalar].text.urlComponentEncoded).iri()` to method

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasHeaderParametersParser.scala
```diff
@@ -76,13 +76,13 @@ case class OasHeaderParameterParser(map: YMap, adopt: Parameter => Unit)(implici
...
     }
-    header.set(ParameterModel.Binding, AmfScalar("header"), Annotations() += SynthesizedField()) // we need to add the binding in order to conform all parameters validations
+    header.set(ParameterModel.Binding, AmfScalar("header"), Annotations.synthesized()) // we need to add the binding in order to conform all parameters validations
     header
   }
 
   protected def parseOas2Header(parameter: Parameter, map: YMap): Unit = {
     val name = Option(parameter.name).map(_.value())
-    parameter.set(ParameterModel.Required, !name.exists(_.endsWith("?")))
+    parameter.set(ParameterModel.Required, AmfScalar(!name.exists(_.endsWith("?"))), Annotations.synthesized())
```
tomsfernandez, line 85: Extract `!name.exists(_.endsWith("?"))` to method

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasLikeEndpointParser.scala
```diff
@@ -141,11 +141,11 @@ abstract class OasEndpointParser(entry: YMapEntry, producer: String => EndPoint,
...
           .as[Seq[YNode]]
-          .map(s => OasLikeSecurityRequirementParser(s, endpoint.withSecurity, idCounter).parse())
-          .collect { case Some(s) => s }
+          .flatMap(s =>
+            OasLikeSecurityRequirementParser(s, (se: SecurityRequirement) => se.adopted(endpoint.id), idCounter)
+              .parse())
 
-        if (securedBy.nonEmpty)
-          endpoint.set(OperationModel.Security, AmfArray(securedBy, Annotations(entry.value)), Annotations(entry))
+        endpoint.set(OperationModel.Security, AmfArray(securedBy, Annotations(entry.value)), Annotations(entry))
```
tomsfernandez, line 148: Shouldn't we check if it is empty? Why omit that?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasLikeOperationParser.scala
```diff
@@ -18,22 +18,22 @@ import amf.plugins.document.webapi.parser.spec.oas.{
...
+abstract class OasLikeOperationParser(entry: YMapEntry, parentId: String)(implicit val ctx: OasLikeWebApiContext)
     extends SpecParserOps {
 
   protected def methodNode: AmfScalar = ScalarNode(entry.key).string()
 
   def parse(): Operation = {
     val operation: Operation = Operation(Annotations(entry))
-    operation.set(Method, methodNode) // add lexical info
-    adopt(operation)
+    operation.set(Method, methodNode, Annotations.inferred()) // add lexical info
```
tomsfernandez: If `// add lexical info` is no longer required as it has already been done, please erase it.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ServerEmitters.scala
```diff
@@ -261,9 +261,14 @@ private case class Servers(default: Option[Server], servers: Seq[Server])
 
 private object Servers {
   def apply(f: FieldEntry): Servers = {
-    val (default, servers) =
-      f.arrayValues(classOf[Server]).partition(_.annotations.find(classOf[SynthesizedField]).isDefined)
-    new Servers(default.headOption, servers)
+    f.arrayValues(classOf[Server]).toList match {
```
tomsfernandez, line 266: Simplify method as complexity has definitely increased

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/OasDocumentParser.scala
```diff
@@ -306,31 +308,32 @@ abstract class OasDocumentParser(root: Root)(implicit val ctx: OasWebApiContext)
...
   private def parseEndpoints(api: WebApi, entry: YMapEntry) = {
     val paths = entry.value.as[YMap]
     val endpoints =
       paths
         .regex("^/.*")
-        .foldLeft(List[EndPoint]())((acc, curr) =>
-          acc ++ ctx.factory.endPointParser(curr, api.withEndPoint, acc).parse())
-    if (endpoints.nonEmpty) api.set(WebApiModel.EndPoints, AmfArray(endpoints), Annotations(entry.value))
+        .foldLeft(List[EndPoint]())((acc, curr) => acc ++ ctx.factory.endPointParser(curr, api.id, acc).parse())
+    api.set(WebApiModel.EndPoints, AmfArray(endpoints, Annotations(entry.value)), Annotations(entry))
```
tomsfernandez, line 339: We shouldn't add the Endpoints field if the `endpoints` map isn't present.

amf-client/shared/src/test/resources/maker/declared-types.json
```diff
@@ -3,9 +3,6 @@
   "info":{
     "title": "test types",
     "description": "empty api only for test types"
-  },
-  "paths": {
```
tomsfernandez: Weird, without this the API is invalid

amf-client/shared/src/test/resources/maker/inherits-declared-types.json
```diff
@@ -3,9 +3,6 @@
   "info":{
     "title": "test types",
     "description": "empty api only for test types"
-  },
-  "paths": {
```
tomsfernandez: Weird, without this the API is invalid

tomsfernandez (changes requested):

amf-client/shared/src/test/scala/amf/maker/DocumentMakerTest.scala
pope1838: This test is not for check the if the example is valid or not, just compare the parsing result against a pure, in memory, document

tomsfernandez: So? Do you need to test invalid APIs in this suite?

pope1838: Not specifically invalid cases, but I want to remove empty endpoints, that now is being parsed and populated in the model, and not makes so mucho sense add an empty list to the fixture, that is used in different cases

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasTypeParser.scala
pope1838: The iri() method applies to all the composed Namespace:
(Namespace.Data + entry.key.as[YScalar].text.urlComponentEncoded)
No so much to reutilize extracting to a method. Already have common methods "+" , urlComponentEncoded and iri()

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasHeaderParametersParser.scala
pope1838: We do this over name only once. Does not make so much sense create a method just for generate the boolean.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/OasLikeEndpointParser.scala
pope1838: If security key is present, field should be set, regarding if there is some value or not.
Not have the key present is not the same that have it without values.
New, the secured by is being set always that the key is present, wherever the list is empty or not.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/domain/ServerEmitters.scala
pope1838: We need to check if there is only one server before make the filter. We could check the length, and them do the filter, or just use the unapply logic of the List class.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/OasDocumentParser.scala
pope1838: This code only can be reach when the key "paths" is present. If paths is present, then we use that entry as source maps of the endpoints field. If path key is present, endpoints field needs to be set. Is not the same have empty paths that not have the key at all.

amf-client/shared/src/test/resources/maker/declared-types.json
pope1838: This file is no used to test validation, just to test parsing. If path is present, the we will have an empty field endpoints that cannot be obtained in a raml api.
As the fixture is the same for deferents specs (on purpose) remove the paths key for the test source is the most convenience solution.

amf-client/shared/src/test/resources/maker/inherits-declared-types.json
pope1838: Same that previous comment.

tomsfernandez (approved):

PR #750, opened 2020-11-20
Fix interfaces WebAPI/AsyncAPI for JS. Added typings

looseale:


amf-client/js/typings/amf-client-js.d.ts
```diff
@@ -1,3 +1,14 @@
+import {
+  CreativeWork,
```
tomsfernandez: This import shouldn't be here

tomsfernandez (changes requested):

tomsfernandez (approved):

PR #760, opened 2020-12-01
APIMF-2625: remove enum values from shape when validating each enum value

AgustinBettati:
validation of a raml type with 70k enum values passes from 2 minutes to 7 seconds.

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/domain/shapes/validation/ShapesNodesValidator.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/domain/shapes/validation/ShapesNodesValidator.scala
@@ -1,6 +1,7 @@
 package amf.plugins.domain.shapes.validation
 
 import amf.core.metamodel.document.PayloadFragmentModel
+import amf.core.metamodel.domain.ShapeModel
 import amf.core.model.document.PayloadFragment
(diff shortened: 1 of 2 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/domain/shapes/validation/ShapesNodesValidator.scala
```diff
@@ -30,14 +31,20 @@ object ShapesNodesValidator {
...
-          s.values.map(v => ValidationCandidate(s, PayloadFragment(v, defaultMediaTypeFor(v))))
-        case _ => Nil
+        case s: Shape if s.values.nonEmpty => shapeEnumCandidates(s)
+        case _                             => Nil
       })
     //call to validation
     PayloadValidationPluginsHandler.validateAll(enumsCandidates.toSeq, severity, env)
   }
 
+  def shapeEnumCandidates(shape: Shape): Seq[ValidationCandidate] = {
```
tomsfernandez: Make private. Should we add something that tests this? I fear someone will edit it and won't realize it is a mistake until a critical issue arrives.

tomsfernandez (changes requested):

tomsfernandez (approved):

PR #799, opened 2021-01-13
APIMF 2732

tomsfernandez:


amf-client/shared/src/test/resources/cycle/raml10/type/json-schema-refs-emission.yaml
```diff
@@ -1,5 +1,7 @@
 type: object
 properties:
-  full-schema: !include external-1.json
-  relative-pointer: !include external-2.json#/definitions/Cat
+  full-schema:
```
tomsfernandez, line 3: It looks like the original now

amf-client/shared/src/test/resources/org/raml/parser/examples/include-json-schema/output.txt.jvm
```diff
@@ -10,8 +10,8 @@ Level: Violation
 /name expected type: String, found: Integer
 
   Level: Violation
-  Target: file://user.json/#/definitions/User/example/bad
-  Property: file://user.json/#/definitions/User/example/bad
+  Target: file://amf-client/shared/src/test/resources/org/raml/parser/examples/include-json-schema/input.raml#/declarations/types/User/example/bad
```
tomsfernandez, line 13: Before, the type User was the external json type with merged examples. Now, the external json type is in the inherits and User is a normal Raml type

amf-client/shared/src/test/resources/upanddown/externals.json
```diff
@@ -50,11 +50,16 @@
             "name": "generated",
             "schema": {
               "type": "object",
-              "properties": {
-                "a": {
-                  "type": "string"
+              "x-amf-merge": [
```
tomsfernandez, line 53: This comes from a RAML type with JSON inheritance. As this was changed to an inheritance, it is emitting that inheritance.

amf-client/shared/src/test/resources/upanddown/cycle/raml10/externals/api.raml.jsonld.raml
```diff
@@ -10,4 +10,5 @@ mediaType: application/json
   post:
     body:
       text/xml: <xs:schema xmlns:xs='http://www.w3.org/2001/XMLSchema' targetNamespace='http://tempuri.org/po.xsd'></xs:schema>
-      application/json: "{ \"type\": \"object\", \"properties\": { \"a\": {\"type\": \"string\"} } }"
+      application/json:
```
tomsfernandez, line 13: Looks like original now

amf-client/shared/src/test/resources/upanddown/cycle/raml10/api-with-multi-line-schema/api.raml.expanded.jsonld
```diff
@@ -52,7 +52,7 @@
                 ],
                 "http://a.ml/vocabularies/document-source-maps#value": [
                   {
-                    "@value": "[(2,0)-(20,0)]"
+                    "@value": "[(2,0)-(19,5)]"
```
tomsfernandez, line 55: some api's lost the final line break in their json schemas when cycling, don't really know why.

amf-client/shared/src/test/resources/org/raml/parser/include/include-jsonschema-typedeclaration-refs/output.txt
```diff
@@ -3,43 +3,41 @@ title: Staff API
...
-            }
-          }
-        },
-        "required": [
-          "firstName",
-          "lastName",
-          "friends"
-        ]
-      }
+    displayName: Example Schema
```
tomsfernandez, line 6: Undesired change but I don't think it is relevant. It appears to lose annotations in inheritance resolving in the ShapeCanonizer. Debugging it is a mess.

We should discuss wether this is relevant or not.

amf-client/shared/src/test/resources/upanddown/cycle/raml10/api-with-multi-line-schema/dumped.raml
```diff
@@ -1,7 +1,7 @@
 #%RAML 1.0
 title: api
 types:
-  mySchema: |
+  mySchema: |-
```
tomsfernandez, line 4: Due to removing line breaks

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/emitters/raml/RamlExternalSchemaWrapperEmitter.scala
```diff
@@ -0,0 +1,66 @@
...
+      }
+      emitExamples(shape, result, ordering,references)
+      result ++= shape.inherits.headOption.toSeq.flatMap(s => Raml10TypeEmitter(s, ordering, ignored, references, forceEntry).entries())
+      b.obj(traverse(ordering.sorted(result), _))
+    } else {
+      shape.inherits.headOption.foreach(s => emitReference(s, b))
+    }
+  }
+
+  private def emitReference(shape: Shape, b: PartBuilder): Unit = shape match {
```
tomsfernandez, line 48: I could extract a "RamlReferenceEmitter" from the Raml10TypeEmitter and use it here.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/external/raml/RamlJsonSchemaExpression.scala
```diff
@@ -32,37 +32,64 @@ case class RamlJsonSchemaExpression(key: YNode,
...
+        val dataNodeResult = NodeDataNodeParser(entry.value, wrapper.id, quiet = false).parse()
+        wrapper.setDefaultStrValue(entry)
+        dataNodeResult.dataNode.foreach { dataNode =>
+          wrapper.set(ShapeModel.Default, dataNode, Annotations(entry))
+        }
+      }
+    )
+    parseExamples(wrapper, value.as[YMap])
+    wrapperName(key).foreach(_ => wrapper.withName(_, Annotations(key)))
+    val typeEntryAnnotations = map.key("type").orElse(map.key("schema")).map(e => Annotations(e)).getOrElse(Annotations())
```
tomsfernandez, line 66: Annotations for `inherits` field

amf-client/shared/src/test/resources/org/raml/parser/include/include-jsonschema-typedeclaration-refs/output.txt
pope1838: Haven't a json schema emitter after resolution makes perfect sense to me.
After resolve the shape, is not the same any more. In other cases, the inheritance could be lost. (Same happen here).

amf-client/shared/src/test/resources/upanddown/cycle/raml10/api-with-multi-line-schema/api.raml.expanded.jsonld
pope1838: Probable takes the Ymap as range instead of the YNode container. Is not relevant at this point.

amf-client/shared/src/test/resources/upanddown/cycle/raml10/api-with-multi-line-schema/dumped.raml
pope1838: Is equivalent, I don't see a big issue here neither.

amf-client/shared/src/test/resources/upanddown/externals.json
pope1838: AMF is the only one that knows how to read a "x-amf-" extension. We are changing the meaning of the shape for any other processor, right?

tomsfernandez: Yes, the thing is that this is a RAML to OAS cycle, a deprecated (?) use case. It is reflecting the AST change. If I resolve the API with CompatibilityPipeline then the output is the same as it was before this change.

pope1838: Great! if it happens only when converting in a wrong way, then should not be a problem.

pope1838 (approved):

PR #800, opened 2021-01-13
APIMF-2759: using new AstIndex with performance improvements

AgustinBettati:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/AstIndexBuilder.scala
```diff
@@ -28,23 +37,25 @@ object AstIndexBuilder {
...
 
 case class AstIndexBuilder private (private val refCounter: AliasCounter)(implicit ctx: WebApiContext) {
 
-  def build(node: YNode, scopes: Seq[ResolutionScope]): AstIndex = {
+  def build(node: YNode, scopes: Seq[ResolutionScope], resolvers: Seq[ReferenceResolver]): AstIndex = {
     val acc = mutable.Map.empty[String, YMapEntryLike]
     scopes.foreach(_.resolve("", YMapEntryLike(node), acc))
     index(YMapEntryLike(node), scopes, acc)
-    AstIndex(acc)
+    AstIndex(acc, resolvers)
```
tomsfernandez, line 53: to avoid passing resolvers as parameter to function, we could create an AstIndex with no resolvers and then `copy()` the AstIndex as it is a case class with the resolvers we want to add.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/ReferenceResolver.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.annotation.tailrec
+
+trait ReferenceResolver {
+  def resolve(reference: String, index: Map[String, YMapEntryLike]): Option[YMapEntryLike]
+}
+
+object FragmentTraversingResolver extends ReferenceResolver {
+
+  override def resolve(reference: String, index: Map[String, YMapEntryLike]): Option[YMapEntryLike] =
+    Try(new URI(reference)).toOption.flatMap { uri =>
```
tomsfernandez, line 17: I remember I tech-talk with Nacho Manzano where he talked about some performance caveats with scala `Try` instead of try catch. Don't remember concretely what the problem was though.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/ReferenceResolver.scala
```diff
@@ -0,0 +1,45 @@
...
+      val baseNode = index.get(baseUri)
+      baseNode.flatMap { node =>
+        val keys = keysOfFragment(fragment)
+        keys.flatMap(traverseFragmentKeys(node, _))
+      }
+    }
+
+  @tailrec
+  private def traverseFragmentKeys(entry: YMapEntryLike, keys: List[String]): Option[YMapEntryLike] =
+    keys.headOption match {
```
tomsfernandez, line 29: Why don't you take into account Sequences? This traversing sounds similar to the AstIndexBuilder's one.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/ReferenceResolver.scala
```diff
@@ -0,0 +1,45 @@
...
+        newNode match {
+          case Some(entry) => traverseFragmentKeys(YMapEntryLike(entry), keys.tail)
+          case None        => None
+        }
+      case None => Some(entry)
+    }
+
+  private def keysOfFragment(fragment: String): Option[List[String]] = {
+    if (fragment.isEmpty) Some(Nil)
+    else if (fragment.startsWith("/")) Some(fragment.stripPrefix("/").split("/").toList)
```
tomsfernandez, line 55: We should centralize this opinionated entry storage mechanism in the AstIndex or somewhere

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/AstIndex.scala
```diff
@@ -1,18 +1,23 @@
...
   def getNode(reference: String): Option[YMapEntryLike] = {
     val toLookUp = clean(reference)
-    map.get(toLookUp)
+    map.get(toLookUp).orElse { callResolvers(toLookUp) }
   }
 
+  private def callResolvers(reference: String): Option[YMapEntryLike] =
+    resolvers.iterator.map(_.resolve(reference, map.toMap)).collectFirst {
+      case Some(entry) =>
+        map.put(reference, entry)
```
tomsfernandez, line 17: isnt this `map` reference an immutable Map?

tomsfernandez: Maybe we should add a separate unit test for this "lazy building"

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/contexts/WebApiContext.scala
```diff
@@ -50,9 +55,21 @@ abstract class WebApiContext(val loc: String,
 
   var jsonSchemaRefGuide: JsonSchemaRefGuide = JsonSchemaRefGuide(loc, refs)(this)
 
+  var indexCache: mutable.Map[String, AstIndex] = mutable.Map[String, AstIndex]()
+
   def setJsonSchemaAST(value: YNode): Unit = {
+    val location = value.sourceName
```
tomsfernandez, line 61: For me the best option is that the Fragment holds it's index. This has several problems though:
- The cache and index would both have to have concurrency control.
- A JSON File can be interpreted using several schemas.

Although this improves performance by saving indexing several times when document A parses a type in Document B we still have unnecesary processing.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/JsonSchemaParser.scala
```diff
@@ -68,7 +74,14 @@ class JsonSchemaParser{
...
     val inheritedDeclarations = getInheritedDeclarations(parentContext)
 
-    new JsonSchemaWebApiContext(url, document.references, cleanNested, inheritedDeclarations, options, parentContext.defaultSchemaVersion)
+    val schemaContext = new JsonSchemaWebApiContext(url,
+                                                    document.references,
+                                                    cleanNested,
+                                                    inheritedDeclarations,
+                                                    options,
+                                                    parentContext.defaultSchemaVersion)
+    schemaContext.indexCache = parentContext.indexCache
```
tomsfernandez, line 83: As you said, this is awkward

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/ReferenceResolver.scala
AgustinBettati: Totally missed this, fixed.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/AstIndex.scala
AgustinBettati: Agree, added specific test for this case.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/jsonschema/AstIndexBuilder.scala
AgustinBettati: I believe passing the parameter of resolvers explicitly might be helpful to understand which elements are needed for the creation of the AstIndex.

tomsfernandez (approved):

PR #806, opened 2021-01-20
APIMF-2241

tomsfernandez:


amf-client/shared/src/test/scala/amf/client/validation/ClientPayloadValidationTest.scala
```diff
@@ -227,5 +227,14 @@ trait ClientPayloadValidationTest extends AsyncFunSuite with NativeOps with Matc
     }
   }
 
+  test("Long type with int64 format is validated as long") {
+    amf.Core.init().asFuture.flatMap { _ =>
+      amf.Core.registerPlugin(PayloadValidatorPlugin)
+      val shape = new ScalarShape().withDataType(DataTypes.Long).withFormat("int64")
```
AgustinBettati, line 233: Will this change have the same impact for other formats such as int32, int, long, int16, int8?

tomsfernandez: This will just impact int64 and long. They are aliases

tomsfernandez: int, int8, int16 and int32 are all parsed to xsd integer type

looseale (approved):

PR #807, opened 2021-01-21
APIMF-2763: use common function for resolving url relative to certain base

AgustinBettati:
depends on https://github.com/aml-org/amf-core/pull/152

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/contexts/WebApiContext.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/contexts/WebApiContext.scala
@@ -4,10 +4,10 @@ import amf.core.client.ParsingOptions
 import amf.core.model.document.{ExternalFragment, Fragment, RecursiveUnit}
 import amf.core.model.domain.Shape
 import amf.core.parser.{Annotations, ParsedReference, ParserContext}
-import amf.core.remote.Context.resolveRelativeTo
 import amf.core.remote._
 import amf.core.unsafe.PlatformSecrets
-import amf.core.utils.{AliasCounter, IdCounter}
+import amf.core.utils.UriUtils.resolve
+import amf.core.utils.{Absolute, AliasCounter, IdCounter, RelativeToIncludedFile}
 import amf.plugins.document.webapi.annotations.DeclarationKey
 import amf.plugins.document.webapi.contexts.parser.oas.OasWebApiContext
 import amf.plugins.document.webapi.parser.spec._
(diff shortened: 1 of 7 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasRefParser.scala
```diff
@@ -157,12 +157,11 @@ class OasRefParser(map: YMap,
     }
 
   private def searchRemoteJsonSchema(ref: String, text: String, e: YMapEntry) = {
-    val fullRef = ctx.resolvedPath(ctx.rootContextDocument, ref)
-    ctx.findJsonSchema(fullRef) match {
-      case Some(u: UnresolvedShape) => copyUnresolvedShape(ref, fullRef, e, u)
+    val fileUrl = ctx.resolvedPath(ctx.rootContextDocument, ref)
```
tomsfernandez: Isn't this the fileUrl + fragment of the json entry?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/external/raml/RamlJsonSchemaExpression.scala
```diff
@@ -291,7 +303,7 @@ case class RamlJsonSchemaExpression(key: YNode,
     }
   }
 
-  private def normalizePath(rawPath: String): String = {
+  private def stripPointsAndFragment(rawPath: String): String = {
```
tomsfernandez, line 306: Why are we doing this?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/contexts/WebApiContext.scala
```diff
@@ -122,20 +124,13 @@ abstract class WebApiContext(val loc: String,
...
-    else if (shouldSkipDot(base, str)) basePath(base) + str.substring(2)
-    else platform.normalizePath(basePath(base).urlDecoded + str)
-
-  def shouldSkipDot(base: String, path: String): Boolean =
-    (basePath(base) == "file://" || basePath(base) == "file:///") && path.startsWith("./")
-
-  def basePath(path: String): String = {
-    val withoutHash = if (path.contains("#")) path.split("#").head else path
-    withoutHash.splitAt(withoutHash.lastIndexOf("/"))._1 + "/"
+  def resolvedPath(current: String, url: String): String = {
```
tomsfernandez: Should this be in another class and be unit tested separately?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/external/raml/RamlJsonSchemaExpression.scala
AgustinBettati: Appears to be coming for here: https://www.mulesoft.org/jira/browse/APIMF-1840. The raw path cannot be simply resolved against the current file location because is can be an absolute reference (only applies for raml).

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasRefParser.scala
AgustinBettati: true, will adjust naming to make it clear.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/contexts/WebApiContext.scala
AgustinBettati: done.

tomsfernandez (approved):

PR #814, opened 2021-01-28
APIMF-2814: maintain precision when validating large numbers

AgustinBettati:
- update everit version to 1.12.2, and org.json to 20201115
 - adjust JSONTokenerHack and JSONObject to new changes in the json library
 - requires changes in ValueEmitter https://github.com/aml-org/amf-core/pull/158

### DIFF
--- a/amf-client/shared/src/test/resources/org/raml/parser/types/builtin/number-float/output.txt
+++ b/amf-client/shared/src/test/resources/org/raml/parser/types/builtin/number-float/output.txt
@@ -4,6 +4,6 @@ types:
   MyCustomType:
     type: number
     minimum: 2.5
-    maximum: 9.3
+    maximum: 9.30
     multipleOf: 3.4
     example: 6.8
--- a/amf-webapi/jvm/src/main/scala/amf/plugins/document/webapi/validation/json/JSONTokenerHack.scala
+++ b/amf-webapi/jvm/src/main/scala/amf/plugins/document/webapi/validation/json/JSONTokenerHack.scala
@@ -12,7 +12,7 @@ class JSONTokenerHack(text: String) extends JSONTokener(text) {
   /** This is extracted from JSONTokener class to replace the JSONObject.stringToValue method used in the superclass
     * so that we can fail when an unquoted value is processed (as all the other json parsers).
     */
-  private def nextValueHack(): Object = hack {
+  private def nextValueHack(): Object = hackDecimal {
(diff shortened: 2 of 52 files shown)


amf-webapi/jvm/src/main/scala/amf/plugins/document/webapi/validation/json/JSONTokenerHack.scala
```diff
@@ -34,66 +33,28 @@ class JSONTokenerHack(text: String) extends JSONTokener(text) {
...
   }
 
   protected def shouldContinueParsing(newChar: Char) = newChar >= ' ' && ",:]}/\\\"[{;=#".indexOf(newChar) < 0
 
-  private def checkNumber(s: String): Option[Object] =
-    try {
-      if (isDecimalNotation(s) || isBiggerThanLong(s)) checkDouble(s) else checkLong(s)
-    } catch {
-      case _: Exception => None
+  private def hack(value: Object) =
```
looseale: Add a more descriptive method name and a description if possible (why we do this hack)

amf-client/shared/src/test/resources/org/raml/parser/types/builtin/number-float/output.txt
```diff
@@ -4,6 +4,6 @@ types:
   MyCustomType:
     type: number
     minimum: 2.5
-    maximum: 9.3
+    maximum: 9.30
```
looseale, line 7: Why it adds a 0 now? As I could see, the method to avoid this in the BaseEmitters is still there

amf-webapi/jvm/src/main/scala/amf/plugins/document/webapi/validation/json/JSONTokenerHack.scala
AgustinBettati: adjusted 👍

amf-client/shared/src/test/resources/org/raml/parser/types/builtin/number-float/output.txt
AgustinBettati: 9.30 is actually the original value, previously it was removed when calling .toDouble. It is now being maintained because BigDecimal.toString conserves zeros to the right. This can be adjusted easily with .stripTrailingZeros(), but i don't think there is a need in removing them right?

looseale (approved):

PR #820, opened 2021-02-03
APIMF-2678: adopt common test artifact

AgustinBettati:


AgustinBettati:
new artifact is present in https://github.com/aml-org/scala-common-test

nschejtman (approved):

PR #836, opened 2021-02-10
Adjust iri for bindings field in (Server,Channel,Operation,Message)BindingsModel

AgustinBettati:


tomsfernandez:
`bindings` can be easily mistaken with `binding`. Why not `bindingSet` or something of the sort?

AgustinBettati:
> `bindings` can be easily mistaken with `binding`. Why not `bindingSet` or something of the sort?

it's a valid point, but the naming seems kind of awkward

pope1838 (dismissed):

pope1838 (approved):

PR #844, opened 2021-02-22
SecurityRequirement as NamedDomainElement

agusayerza:


tomsfernandez (changes requested):
Why protected?

agusayerza:
> Why protected?

Resolved. Autogenerated by Intellij as parent class `NamedDomainElement` defines `protected def nameField`

amf-webapi/shared/src/main/scala/amf/plugins/domain/webapi/models/security/SecurityRequirement.scala
```diff
@@ -1,20 +1,18 @@
...
 import amf.plugins.domain.webapi.metamodel.security.SecurityRequirementModel
 import amf.plugins.domain.webapi.metamodel.security.SecurityRequirementModel._
 import org.yaml.model.YPart
-import amf.core.utils.AmfStrings
 
-case class SecurityRequirement(fields: Fields, annotations: Annotations) extends DomainElement {
+case class SecurityRequirement(fields: Fields, annotations: Annotations) extends NamedDomainElement {
 
-  def name: StrField                           = fields.field(Name)
+  override def nameField: Field                = Name
```
tomsfernandez, line 13: why change name for nameField?

tomsfernandez (changes requested):

tomsfernandez (approved):

PR #845, opened 2021-02-23
ALS-1224: Security schemes name annotations

agusayerza:


amf-webapi/shared/src/main/scala/amf/plugins/domain/webapi/models/security/ParametrizedSecurityScheme.scala
```diff
@@ -1,24 +1,23 @@
...
 case class ParametrizedSecurityScheme(fields: Fields, annotations: Annotations)
-    extends DomainElement
+    extends NamedDomainElement
     with WithSettings {
 
-  def name: StrField         = fields.field(Name)
-  def description: StrField  = fields.field(Description)
-  def scheme: SecurityScheme = fields.field(Scheme)
-  def settings: Settings     = fields.field(SettingsField)
+  override def nameField: Field = Name
```
tomsfernandez: should be private

tomsfernandez (changes requested):

amf-webapi/shared/src/main/scala/amf/plugins/domain/webapi/models/security/ParametrizedSecurityScheme.scala
tomsfernandez: or protected

tomsfernandez (changes requested):

tomsfernandez (approved):

PR #868, opened 2021-03-11
parsing plugins remodularization

AgustinBettati:
This PR includes:
- APIMF-2877: definition of parsing plugins and adoption in AMFCompiler
- APIMF-2905: remove validation methods from parse plugin interface
- APIMF-2907: create default environments with parsing plugins
- APIMF-2880: define interface for fallback mechanism when no plugin is matched

depends on https://github.com/aml-org/amf-aml/pull/198 and https://github.com/aml-org/amf-core/pull/178


amf-client/shared/src/main/scala/amf/client/commands/CommandHelper.scala
```diff
@@ -28,6 +39,7 @@ trait CommandHelper {
     amf.core.AMF.registerPlugin(Raml08Plugin)
     amf.core.AMF.registerPlugin(Oas20Plugin)
     amf.core.AMF.registerPlugin(Oas30Plugin)
```
tomsfernandez, line 29: [OOS] - Not part of REMOD, could be taken to another commit.

amf-client/shared/src/test/resources/validations/reports/model/invalid-reference-defined-in-lib.report
```diff
@@ -0,0 +1,38 @@
+Model: file://amf-client/shared/src/test/resources/validations/invalid-reference/error-in-lib.raml
+Profile: RAML 1.0
+Conforms? false
+Number of results: 4
```
tomsfernandez, line 4: Why is there a new test?

amf-client/shared/src/test/scala/amf/cycle/OasRecursiveFilesCycleTest.scala
```diff
@@ -43,7 +42,7 @@ class OasRecursiveFilesCycleTest extends FunSuiteCycleTests {
 
       RuntimeCompiler
         .forContext(
-          compilerContextBuilder.withEnvironment(Environment.apply()).build(),
+          compilerContextBuilder.build(),
```
tomsfernandez, line 45: Can we break this interface?

amf-client/shared/src/test/scala/amf/error/RamlParserErrorTest.scala
```diff
@@ -98,19 +98,15 @@ class RamlParserErrorTest extends ParserErrorTest {
   test("Inline external fragment from non mutable ref") {
     validate(
       "error/inline-non-mutable-ref/api.raml",
-      badInclude1 => {
-        badInclude1.level should be("Violation")
-        badInclude1.message should startWith("Fragments must be imported by using '!include'")
-      },
-      badInclude2 => {
```
tomsfernandez, line 105: Why do we have one less assertion?

amf-webapi/shared/src/main/scala/amf/client/environment/ApiEnvironment.scala
```diff
@@ -0,0 +1,30 @@
+package amf.client.environment
+
+import amf.client.remod.AMFEnvironment
+import amf.plugins.document.graph.AMFGraphParsePlugin
+import amf.plugins.document.webapi._
+
+// defined private unit mayor version is released
+private[amf] object ApiEnvironment {
```
tomsfernandez, line 8: Hmm suspicious. With this we are forcing adopters to depend on all of the artifacts we support. I agree that we have to make it easy for adopters to migrate but still, suspicious.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/AsyncPlugin.scala
```diff
@@ -35,10 +35,7 @@ sealed trait AsyncPlugin extends OasLikePlugin {
               wrapped: ParserContext,
               ds: Option[AsyncWebApiDeclarations] = None): AsyncWebApiContext
 
-  override def parse(document: Root,
-                     parentContext: ParserContext,
-                     platform: Platform,
```
tomsfernandez, line 40: I saw Platform being used as a parameter in several places, why do we have this if Platform is a Singleton for the distribution? Is this usage legacy?

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/references/RamlReferenceHandler.scala
```diff
@@ -0,0 +1,95 @@
...
+  /** Update parsed reference if needed. */
+  override def update(reference: ParsedReference, compilerContext: CompilerContext)(
+      implicit executionContext: ExecutionContext): Future[ParsedReference] = {
+    if (reference.isExternalFragment)
+      handleRamlExternalFragment(reference, compilerContext)
+    else Future.successful(reference)
+  }
+
+  private def handleRamlExternalFragment(reference: ParsedReference, compilerContext: CompilerContext)(
+      implicit executionContext: ExecutionContext): Future[ParsedReference] = {
```
tomsfernandez, line 31: [OOS] - A nice cleanup could be done here.

tomsfernandez (dismissed):

amf-client/shared/src/test/resources/validations/reports/model/invalid-reference-defined-in-lib.report
AgustinBettati: This was included when moving verifyValidReferenceKind from the document plugin interface to the raml plugin, i found a specific case that was not being tested so i included this test case (using invalid references within in a library).

amf-client/shared/src/test/scala/amf/cycle/OasRecursiveFilesCycleTest.scala
AgustinBettati: This interface has not been broken (CompilerContextBuilder), removed this code as it was useless.

amf-client/shared/src/test/scala/amf/error/RamlParserErrorTest.scala
AgustinBettati: This was due to the change of verifyValidReferenceKind. This test case has a an external fragment that references a raml library incorrectly. This incorrect reference used to be validated with the raml plugin (even though the file in not raml related), while now as these reference validation are specific for raml it is no longer present.
We analyzed this behavior change with @pope1838 and agreed that the impact is very reduced and can be considered correct.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/AsyncPlugin.scala
AgustinBettati: Introduced this changed as it was a parameter that was unused in all plugin implementations. As you mention, this most likely was a legacy usage.

amf-webapi/shared/src/main/scala/amf/client/parse/Oas20Parser.scala
```diff
@@ -3,7 +3,7 @@ package amf.client.parse
 import amf.client.environment.Environment
 import amf.core.registries.AMFPluginsRegistry
 import amf.core.remote.Oas20
-import amf.plugins.document.webapi.Oas20Plugin
+import amf.plugins.document.webapi.{Async20ParsePlugin, ExternalJsonYamlRefsParsePlugin, Oas20ParsePlugin, Oas20Plugin}
```
looseale: Delete import of Async20ParsePlugin, also ExternalJsonYamlRefsParsePlugin.
In general all parsers looks to have imports of plugins that are not used (were added in a previous commit but the uses were deleted in a newer one)

looseale (dismissed):

looseale (approved):

PR #871, opened 2021-03-16
APIMF-2936: added cross spec restriction check. Moved it out of AMFCompiler

tomsfernandez:
Depends on: https://github.com/aml-org/amf-core/pull/179﻿

### DIFF
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/CrossSpecRestriction.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/CrossSpecRestriction.scala
@@ -1,0 +1,33 @@
+package amf.plugins.document.webapi
+
+import amf.core.Root
+import amf.core.errorhandling.ErrorHandler
+import amf.core.parser.{ParserContext, Reference}
+import amf.core.remote.Vendor
+import amf.plugins.features.validation.CoreValidations.InvalidCrossSpec
+
+trait CrossSpecRestriction {
+
+  protected def vendors: Seq[String]
+  protected def validVendorsToReference: Seq[Vendor] = Seq.empty
+
+  // TODO: all documents should have a Vendor
+  protected def restrictCrossSpecReferences(optionalReferencedVendor: Option[Vendor], reference: Reference)(
+      implicit errorHandler: ErrorHandler): Unit = {
+    val possibleReferencedVendors = vendors ++ validVendorsToReference.map(_.name)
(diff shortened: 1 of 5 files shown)


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/CrossSpecRestriction.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  protected def vendors: Seq[String]
+  protected def validVendorsToReference: Seq[Vendor] = Seq.empty
+
+  // TODO: all documents should have a Vendor
+  protected def restrictCrossSpecReferences(optionalReferencedVendor: Option[Vendor], reference: Reference)(
+      implicit errorHandler: ErrorHandler): Unit = {
+    val possibleReferencedVendors = vendors ++ validVendorsToReference.map(_.name)
+    optionalReferencedVendor.foreach { referencedVendor =>
+      if (!possibleReferencedVendors.contains(referencedVendor.name)) {
```
AgustinBettati, line 19: There used to be a check that allowed json references for all cases, im in favor of removing this custom check but do all plugins have this vendor set in the validVendorsToReference?

tomsfernandez: ExternalFragments doesn't have a vendor so it wouldn't apply for this condition. 

I think that the previous condition stated that an ExternalFragment could reference another document from another plugin. It was a hack for a cyclic reference case where ALS sent the ExternalFragment to parse and as it had a $ref it followed the parsing path. (I think)

looseale (dismissed):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/CrossSpecRestriction.scala
looseale: It cames from this issue https://www.mulesoft.org/jira/browse/APIMF-2143 and it doesn't added a test. So we should check it to be sure that doesn't break anything.

tomsfernandez: Yup it has tests. 

See: https://github.com/aml-org/amf/blob/9ff718f5fb5418fe8ce6afa45f16399d7aeaf8ca/amf-client/shared/src/test/scala/amf/cycle/OasRecursiveFilesCycleTest.scala#L14-L53

looseale (approved):

PR #872, opened 2021-03-18
APIMF 2881

tomsfernandez:
Related to:
- https://github.com/aml-org/amf-core/pull/180
- https://github.com/aml-org/amf-aml/pull/202


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/validation/remod/RemodValidationRunner.scala
```diff
@@ -0,0 +1 @@
+
```
AgustinBettati: Remove empty file

tomsfernandez: Removed

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/validation/PayloadsInApiCollector.scala
```diff
@@ -60,6 +60,8 @@ class PayloadsInApiCollector(model: BaseUnit) {
     }.toSeq
   }
 
+  private def isStrictAnyShape(shape: AnyShape) = shape.meta == AnyShapeModel
```
looseale, line 63: This method should include the fields check, otherwise the name should be "isAnyShape" or something like that

looseale (dismissed):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/validation/PayloadsInApiCollector.scala
tomsfernandez: why fields check? The only way to know if an AnyShape is in fact an AnyShape is to check the instance's metamodel.

AgustinBettati (approved):

PR #873, opened 2021-03-19
Render plugins remodularization

AgustinBettati:
depends on:
https://github.com/aml-org/amf-aml/pull/204
https://github.com/aml-org/amf-core/pull/182


amf-client/shared/src/test/scala/amf/validation/ValidationProfilesCycle.scala
```diff
@@ -50,31 +50,31 @@ class ValidationProfilesCycle extends AsyncFunSuite with PlatformSecrets {
...
+    cycle(exampleFile, VocabularyYamlHint, Yaml, Aml).zip(expected).map(checkDiff)
   }
 
   test("Loading and serializing validations with union type") {
     val expectedFile             = "validation_profile_example_gold.yaml"
     val exampleFile              = "validation_profile_example.yaml"
     val validation               = Validation(platform)
     val expected: Future[String] = platform.resolve(basePath + expectedFile).map(_.stream.toString)
-    cycle(exampleFile, VocabularyYamlHint, Yaml, Raml).zip(expected).map(checkDiff)
+    cycle(exampleFile, VocabularyYamlHint, Yaml, Aml).zip(expected).map(checkDiff)
```
tomsfernandez, line 77: Weird change

tomsfernandez:
This PR contains:

- Removed literals in tests for constants
- Added tests
- Lots of formatting
- Removed error handler from RenderOptions and is now passed explicitly as parameter
- Refactored default APIEnvironment to use the withPlugins
- Added adapters to current plugins
- Removed ShapeRenderOptions from method signature and added to RenderOptions

If they are not 100% needed I would probably move the test changes to a different commit.

**(EDIT)**: I think it's great that the error handler is now explicitly passed as parameter. Makes the interface much more meaningfull.

amf-client/shared/src/test/scala/amf/validation/ValidationProfilesCycle.scala
AgustinBettati: This change was made to fix the test.
This is due a change in how the serializer searches for a plugin for rendering. Previously AMFSerializer would search for a plugin that complies with the vendor, mediaType, and unparse method, and as a fallback it searchs for a plugin that complies with the mediaType and unparse (without taking into account the vendor).
Since this logic is now present in the applies method of the render plugin (currently in AMFRenderPluginAdapter), there is no fallback when the vendor is incorrect.

tomsfernandez (approved):

PR #934, opened 2021-05-03
APIMF-2920: define id in unresolved discriminator mapping shape and avoid violation

AgustinBettati:
depends on https://github.com/aml-org/amf-core/pull/230


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/declaration/OasRefParser.scala
```diff
@@ -128,7 +128,7 @@ class OasRefParser(map: YMap,
       None,
       Some((k: String) => shape.set(LinkableElementModel.TargetId, k)),
       shouldLink = false).withName(declarationNameOrResolvedRef, Annotations()).withSupportsRecursion(true)
-    tmpShape.unresolved(declarationNameOrResolvedRef, e, "warning")(ctx)
```
AgustinBettati, line 131: These warnings where adjusted because they where not working and now they do (unresolved was always violation). Don't want to break any validation behaviour.

nschejtman (approved):

PR #937, opened 2021-05-03
remod breaking shapes

tomsfernandez:
See: https://github.com/aml-org/amf-core/pull/232


looseale (approved):

tomsfernandez:
Merging it because there is something in Jenkins that's making it fail. Works fine with source mode.

PR #939, opened 2021-05-05
Abstract interfaces/remove generic vendors

pope1838:


amf-client/shared/src/test/resources/production/inter-spec-refs/reports/raml08-raml1-datatype.report
```diff
@@ -1,10 +1,18 @@
 Model: file://amf-client/shared/src/test/resources/production/inter-spec-refs/raml08-raml1-datatype/api.raml
 Profile: RAML 1.0
 Conforms? false
-Number of results: 1
+Number of results: 2
 
 Level: Violation
 
+- Source: http://a.ml/vocabularies/amf/parser#closed-shape
```
pope1838, line 8: Consequence of parsing with specific raml 0.8 vendor instead of generic raml

amf-client/shared/src/test/resources/upanddown/somecars.raml
```diff
@@ -127,39 +127,6 @@ mediaType:
 version: 1.0.0
 protocols:
   - https
-/history:
```
pope1838, line 130: Ordening change. Lexical order should never have applied because we are passing from an OAS to a RAML

looseale (approved):

PR #970, opened 2021-05-30
WIP

pope1838:
Depends on:
- https://github.com/aml-org/amf-aml/pull/280

Failing suits:
- UniqueAmfGraphSyamlConversionExceptionTest: changes on ids and profile name (profile comes empty on parsing errors, because are collected at code, any constraint of profile was applied).
- ProductionResolutionTest: big apis that changes on jsonlds flattened goldens (missing source maps).
- InvalidsRefElementTest: Throws file not found, and the files doesn't exists
- OasInterSpecRefsReportTest error changes
- Raml10CycleTestByDirectory: vendor mediatype changed for external fragments (looks ok). New annotations appears (on external fragment would be ok, but other cases are weird).
- MultiAmfGraphSyamlConversionExceptionTest : validation profile is not present at the report, as are parsing reports.
- RamlInterSpecRefsReportTest: the diff or the report is the profile name, looks ok.
- JsonSchemaUniquePlatformUnitValidationsTest: only fails when all the suits are run.

Discussion topics:
- Profile field of the validation report on parsing and transformation results. What we should do? left it empty? use the source vendor matching profile?

### DIFF
--- a/amf-client/js/src/test/scala/amf/convert/JsOutputBuilderTest.scala
+++ b/amf-client/js/src/test/scala/amf/convert/JsOutputBuilderTest.scala
@@ -1,7 +1,8 @@
 package amf.convert
 
+import amf.client.environment.{AsyncAPIConfiguration, WebAPIConfiguration}
+import amf.client.remod.amfcore.config.RenderOptions
 import amf.core.AMFSerializer
-import amf.core.emitter.RenderOptions
 import amf.core.model.document.BaseUnit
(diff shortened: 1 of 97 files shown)


tomsfernandez:
About the profile: we can't use the source vendor as profile as we have two cases that don't match this behaviour:
- AML: it uses the dialect name as Profile
- Custom Validations: I can validate a unit with source vendor RAML 1.0 and profile 'cyclone' and in the report it should have profile 'cyclone'

PR #996, opened 2021-06-10
Fix annotation on referenced example

agusayerza:


amf-client/shared/src/test/resources/upanddown/cycle/raml10/referenced-example/api.raml.jsonld.raml.ignore
```diff
@@ -0,0 +1,14 @@
+#%RAML 1.0
```
tomsfernandez, line 1: Why ignore?

tomsfernandez:
Please cherry-pick into remod-breaking afterwards

amf-client/shared/src/test/resources/upanddown/cycle/raml10/referenced-example/api.raml.jsonld.raml.ignore
agusayerza: "Generate golden from json-ld for referenced-example" and "Parse golden from json-ld for referenced-example" generate different goldens. This bug is independent on this annotation change. Was planning on opening an issue after merge

tomsfernandez (approved):

PR #998, opened 2021-06-10
APIMF 2981 final

tomsfernandez:
- Related to: https://github.com/aml-org/amf-core/pull/270


pope1838 (approved):

amf-api-contract/shared/src/main/scala/amf/plugins/domain/shapes/validation/JsonCandidateValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import amf.ProfileName
+import amf.client.remod.amfcore.plugins.validate.ValidationConfiguration
+import amf.core.validation._
+import org.mulesoft.common.core.CachedFunction
+import org.mulesoft.common.functional.MonadInstances.identityMonad
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object JsonCandidateValidator {
```
AgustinBettati: Found this name confusing as this object is simple facade which is agnostic to specific syntax.

tomsfernandez: Seemed appropiate to me as this validates candidates with "application/json" mediatype

tomsfernandez: changed it for CandidateValidator

PR #1004, opened 2021-06-15
APIMF-3021: fix raml08 parameter not being added to entries in oas request parser

arielmirra:
fixes https://www.mulesoft.org/jira/browse/APIMF-3021


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/Oas20RequestParser.scala
```diff
@@ -67,6 +67,7 @@ case class Oas20RequestParser(map: YMap, adopt: Request => Unit)(implicit ctx: O
       "baseUriParameters".asOasExtension,
       entry => {
         entry.value.as[YMap].entries.headOption.foreach { paramEntry =>
+          entries += paramEntry
```
arielmirra, line 70: Without this, line 93 did a `.head` of an empty list

AgustinBettati (approved):

PR #1029, opened 2021-06-29
APIMF-3211: add map of null key elements Fixes incorrect overriding when there are multiple elements with null key

arielmirra:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtensionDomainElementMerge.scala
```diff
@@ -167,47 +168,49 @@ class ExtensionDomainElementMerge(restrictions: MergingRestrictions,
...
+                               key: KeyField,
+                               master: AmfArray,
+                               extension: AmfArray): Unit = {
 
-    val asSimpleProperty                  = isSimpleProperty(key)
-    var existing: Map[Any, DomainElement] = buildElementByKeyMap(key, master)
-    var nullKey: Option[DomainElement]    = findElementWithNullKey(key, master)
+    val asSimpleProperty                                 = isSimpleProperty(key)
+    var existingElements: Map[Any, DomainElement]        = buildElementByKeyMap(key, master)
+    var existingNullKeyElements: Map[Obj, DomainElement] = findElementsWithNullKey(key, master)
```
nschejtman, line 182: Add comment saying that "if we have multiple elements with null key we merge by `meta obj`"

nschejtman (dismissed):

nschejtman (approved):

PR #1036, opened 2021-07-06
APIMF 3202

tomsfernandez:
- APIMF-3092/0: uncommented UnitCacheTest, deleted JVM/JS pairs for it
- APIMF-3092/1: Deleted AMF Logger
- APIMF-3092/2: Renamed bu for baseUnit
- APIMF-3092/3: Hid constructors in scala client Shape models
- APIMF-3092/4: adapted payload validators to new interface, added helper in ProvidedMediaTypes
- APIMF-3092/5: added method to AMFConfig to be coherent with AMFGraphConfig
- Fixed rebase changes

Related to: https://github.com/aml-org/amf-aml/pull/327


amf-api-contract/shared/src/main/scala/amf/apicontract/client/common/ProvidedMediaType.scala
```diff
@@ -8,6 +8,8 @@ import scala.scalajs.js.annotation.{JSExportAll, JSExportTopLevel}
 @JSExportAll
 object ProvidedMediaType {
 
+  def build(mediaType: String, syntax: String): String = mediaType + syntax
```
AgustinBettati: this will not build a valid mediaType, it is missing the '+' sign.

nschejtman: Isn't this just Vendor + Syntax? Media types already include syntax

nschejtman (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/client/common/ProvidedMediaType.scala
nschejtman: Strictly speaking we should use these media types (not registered in IANA): https://en.wikipedia.org/wiki/Media_type#Unregistered_tree

tomsfernandez: Yup, agree on this one. Will fix it.

tomsfernandez: Dropped method as these media types will no longer be used. Change pending

PR #1049, opened 2021-07-20
APIMF-3212: fix duplication of enum values in raml extension

arielmirra:


amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtendsResolutionStage.scala
```diff
@@ -29,11 +29,11 @@ import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 
 /**
-  * 1) Get a queue of resource types for this Endpoint.
-  * 2) Resolve each resource type and merge each one to the endpoint. Start with the closest to the endpoint.
-  * 3) Get the traits as branches, as described in the spec, to get the order of the traits to apply.
-  * 4) Resolve each trait and merge each one to the operation in the provided order..
-  * 5) Remove 'extends' property from the endpoint and from the operations.
+  *   1. Get a queue of resource types for this Endpoint.
```
AgustinBettati, line 32: not sure if this is intentional.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtensionsResolutionStage.scala
```diff
@@ -317,9 +317,9 @@ class ExtensionResolutionStage(override val profile: ProfileName, override val k
                                           other: AmfArray,
                                           extensionId: String,
                                           extensionLocation: Option[String]): Unit = {
-    other.values.foreach { value =>
-      target.add(field, value)
-    }
+    val targetValues = target.fields.get(field).asInstanceOf[AmfArray].values
```
AgustinBettati, line 320: Are we sure that the value of the `target` element in the specific `field` is always an AmfArray?

AgustinBettati (changes requested):

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtendsResolutionStage.scala
arielmirra: yes, it's the format for scaladoc numbered list, fixed it in case we want to export scaladocs sometime (and now Intellij formatter displays the list correctly instead of inlined)

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/resolution/stages/ExtensionsResolutionStage.scala
arielmirra: yes, that method is called `setDomainElementArrayValue` and always sets an array value. I've just changed the array it sets to not have duplicated values.

AgustinBettati (approved):

PR #1068, opened 2021-08-04
APIMF-3216: added parseContent tests to make sure they don't return ExternalFragments with YAML/JSON contents

tomsfernandez:
Depends on:
- https://github.com/aml-org/amf-core/pull/310


AgustinBettati (commented):
Commit message does not match with issue

Issue #869, opened 2021-03-12
RAML nillable shorthand properties lose default value


kdubb:
Your issue may already be reported! Please search on Github issues before creating one.

* **I'm submitting a ...**
  - [X] bug report
  - [ ] feature request
  - [ ] question
  - [ ] other

* **What is the current behavior?**

RAML nillable shorthand properties lose their default value.

* **If the current behavior is a bug, please provide the minimal steps to reproduce with AMF code. Please don't upload AMF code mixed with yours.**

For the following RAML, the `schema().defaultValue()` property of each `operation.request().headers()` has a default value except for the last parameter `def2`; which uses RAML nillable shorthand.
```raml
/tests:
  get:
    displayName: fetchTest
    headers:
      int:
        type: integer
        default: 1
      def1?:
        type: integer
        default: 5
      def2:
        type: integer?
        default: 10
    responses:
      200:
        body:
          type: Test

```

* **What is the expected behavior?**

The `defaultValue` should be available for all properties.

* **What is the motivation/use case for changing the behavior? (for feature requests)**

* **Please tell us about your environment:**
  
  - AMF Version: 4.7.0
  - AMF Distribution: JVM
  - Operating System: macOS
  - Link to your project: https://github.com/outfoxx/sunday-generator

* **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, Gitter, etc)

kdubb:
Note that in expanded union form... e.g.
```raml
def3:
  type: integer | nil
  default: 20
```

The default value is appropriately assigned.

AgustinBettati:
Hello @kdubb! thanks for bringing this up. 
I created an issue in our internal tracking system for this: https://www.mulesoft.org/jira/browse/APIMF-2973.
I will keep you updated on any progress.

kdubb:
@AgustinBettati I have found **numerous** issues with nillable shorthand properties. I haven't reported all of them because they basically seem to be just unsupported at the current time.

That being said, _please_ fix and support them! RAML is concise and this is one of the things that makes it concise and why we use it exclusively as our design language over OpenAPI.

AgustinBettati:
Trailing question marks in properties and schemas are fully supported, what are exactly the issues you are encountering?

kdubb:
As I said there have been numerous. I'll continue to report them as we re-encounter them.

[...]

PR #1073, opened 2021-08-04
APIMF-2973: make nillable type behave like nil union

arielmirra:
- Make `type: example?` be equivalent to `type: example | nil`
- This is the correct behavior by the [RAML 1.0 spec]((https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#nil-type))
- other facets are now in the union instead of inside the non-nil anyOf

Fixes issues [APIMF-2973](https://www.mulesoft.org/jira/browse/APIMF-2973) corresponding to github issue #869


arielmirra:
misspelled branch name

PR #1090, opened 2021-08-18
APIMF-3025: generate a PropertyShape when `required` facet is inlined

arielmirra:
- add required facet as criteria to detect an object
- generate a PropertyShape when required facet is inlined

### DIFF
--- a/amf-client/shared/src/test/resources/validations/raml-inline-required.raml
+++ b/amf-client/shared/src/test/resources/validations/raml-inline-required.raml
@@ -1,0 +1,28 @@
+#%RAML 1.0
+title: "test api"
+
+types:
+    testSchema: |
+      {
+        "required": [
+          "option_two"
+        ]
+      }
+
+/pets:
+  get:
+    responses:
+      200:
+        description: Not Found
+        body:
+          application/json:
+            type: testSchema
+            examples:
+              plain-object:
+                option_two: {}
+              null-object:
+                option_two: null
+              a-string:
+                option_two: this is a string
+              a-number:
+                option_two: 1
--- a/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/parser/types/InlineOasTypeParser.scala
+++ b/amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/parser/types/InlineOasTypeParser.scala
@@ -46,7 +46,6 @@ import amf.plugins.domain.webapi.models
 import amf.plugins.domain.webapi.models.IriTemplateMapping
 import amf.validations.ParserSideValidations._
 import org.yaml.model._
-import amf.core.utils.AmfStrings
 
 import scala.collection.mutable
 import scala.util.Try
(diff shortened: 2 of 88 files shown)


amf-client/shared/src/test/resources/validations/raml-inline-required.raml
```diff
@@ -0,0 +1,21 @@
...
+/pets:
+  get:
+    responses:
+      200:
+        description: Not Found
+        body:
+          application/json:
+            type: testSchema
+            example:
+              option_two: {}
```
AgustinBettati: We should add more valid examples here where the property option_two is defined with different types such as a string, array, number, and already defined object. This will ensure that setting the property range with an AnyShape is working as expected.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/parser/types/InlineOasTypeParser.scala
```diff
@@ -725,6 +725,18 @@ case class InlineOasTypeParser(entryOrNode: YMapEntryLike,
     }
   }
 
+  private def generateUndefinedRequiredProperties(requiredFields: Map[String, YNode], shape: NodeShape): Unit = {
+    val presentProperties = shape.fields.fields().map(_.field.value.name).toSeq
+    requiredFields.keys
+      .filter(!presentProperties.contains(_))
+      .foreach(requiredProperty => {
+        shape
+          .withProperty(requiredProperty)
```
AgustinBettati: I think that for this case both the PropertyShape and the AnyShape should be annotated as VirtualElements, not sure if your already checked this with pope.

AgustinBettati (dismissed):

amf-client/shared/src/test/resources/validations/raml-inline-required.raml
AgustinBettati: Would also consider trying the value null.

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/parser/types/InlineOasTypeParser.scala
```diff
@@ -909,7 +922,7 @@ case class InlineOasTypeParser(entryOrNode: YMapEntryLike,
                 val required = ScalarNode(entry.value).boolean().value.asInstanceOf[Boolean]
                 property.set(
                   PropertyShapeModel.MinCount,
-                  AmfScalar(if (required) 1 else 0, Annotations.synthesized()), // does this annotation make sense? should parent not be `synthetized`?
+                  AmfScalar(if (required) 1 else 0, Annotations.synthesized()), // does this annotation make sense? should parent not be `synthesized`?
                   Annotations(entry) += ExplicitField()
```
pope1838: El field deberia ser synthesized, no el value. Porque se marca como explicito? es una contradiccion

amf-webapi/shared/src/main/scala/amf/plugins/document/webapi/parser/spec/oas/parser/types/InlineOasTypeParser.scala
```diff
@@ -699,17 +699,16 @@ case class InlineOasTypeParser(entryOrNode: YMapEntryLike,
...
         Annotations(pe.value)
       } orElse {
         patternPropEntry.map { pp =>
           Annotations(pp.value)
         }
-      } getOrElse { Annotations() }
+      } getOrElse { Annotations.virtual() }
 
       if (properties.nonEmpty)
         shape.set(NodeShapeModel.Properties, AmfArray(properties.values.toSeq, valueAnnotations), valueAnnotations)
```
pope1838: Aca usas la misma annotation para el value y para el field.
Value annotation esta bien, porque si tenes el entry, la creas con entry.value, y si no es virtual.
La del field esta mal, porque mandas value, pero en realidad ahi es o bien de todo el entry como "Annotations(propertyEntry)" o bien Annotations.inferred()

pope1838 (approved):

PR #1105, opened 2021-08-26
chore: disable api query build step as it's not being used

arielmirra:


tomsfernandez:
Closed in favor of #1106

Issue #901, opened 2021-04-13
Incompatible types [array, array] when overriding an object property with an array of array


deiteris:
Your issue may already be reported! Please search on Github issues before creating one.

* **I'm submitting a ...**
  - [ ] bug report
  - [ ] feature request
  - [x] question
  - [ ] other

* **What is the current behavior?**

The following structure of types produces the `Incompatible types [array, array]` error while producing a correct output:
```yaml
types:
  BaseObject:
    type: object
    properties:
      items:
        type: array
        items:
          type: any
  
  ArrayOfObjects:
    type: object[]

  CustomObjectList:
    type: BaseObject
    properties:
      items:
        type: array
        items:
          type: ArrayOfObjects
```

But, if we change the type of the `items` in array of the `BaseObject` to `array` type, the validation passes.

According to RAML 1.0 spec, it is allowed to override a parent type with narrower subtype (and `array` of `array` should be narrower than `array` of `any`), but it seems to fail in this particular case.
https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#object-type-specialization
>2) the type declaration of a defined property in the parent type can only be changed to a narrower type (a specialization of the parent type) in the sub-type.

So the question is: Is there something I'm missing or is it a bug?

* **What is the expected behavior?**

Validation should pass.

* **What is the motivation/use case for changing the behavior? (for feature requests)**

* **Please tell us about your environment:**
  
  - AMF Version: 4.7.1 / 4.7.2 / 4.8.0-SNAPSHOT
  - AMF Distribution: JVM and JS 
  - Operating System: Windows 10H2 / CentOS Linux release 8.3.2011

* **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, Gitter, etc)

tomsfernandez:
Hi @deiteris. This is a bug. The internal ticket issue is: https://www.mulesoft.org/jira/browse/APIMF-3032. Thanks!

arielmirra:
Hi @deiteris! This bug has been fixed and will be present in the next AMF release on Sep 21st, both in AMF 4 (which will enter maintenance mode) and AMF 5 (which is in beta and will go live by that date). Thank you!

PR #1126, opened 2021-09-07
APIMF-3032: fix when array overrides type any

arielmirra:
fixes github issue #901


amf-webapi/shared/src/main/scala/amf/plugins/domain/shapes/resolution/stages/shape_normalization/MinShapeAlgorithm.scala
```diff
@@ -84,6 +84,11 @@ private[stages] class MinShapeAlgorithm()(implicit val context: NormalizationCon
         case baseArray: MatrixShape if superShape.isInstanceOf[MatrixShape] =>
           val superArray = superShape.asInstanceOf[MatrixShape]
           computeMinMatrix(baseArray, superArray)
+        case baseArray: MatrixShape
+            if superShape.isInstanceOf[ArrayShape] &&
```
tomsfernandez: extract to function

arielmirra: done 👌🏽

nschejtman (approved):

PR #1130, opened 2021-09-10
APIMF-3356: add resolution step for headers in components

arielmirra:
- headers declared in components in OAS were being named by the key of the reference rather than the 'parameterName' field


arielmirra:
Closing it as we resolved this is a won't do.

PR #1179, opened 2021-10-22
APIMF-3402: check for ref name or assign 'schema'

arielmirra:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/oas/parser/domain/OasContentParser.scala
```diff
@@ -53,8 +58,9 @@ case class OasContentParser(entry: YMapEntry, producer: Option[String] => Payloa
     map.key(
       "schema",
       entry => {
-        OasTypeParser(entry, shape => shape.withName("schema"))(
-          WebApiShapeParserContextAdapter(ctx))
+        val hasRef  = entry.value.value.asInstanceOf[YMap].key("$ref")
+        val refName = if (hasRef.isDefined) getReferenceName(hasRef.get) else "schema"
```
AgustinBettati: We cound extract this to a method with a name that makes the operation explicit, something like createNameFromRefUrl

AgustinBettati (approved):

PR #1187, opened 2021-10-29
APIMF-3499: fix merging type error by avoiding uriParameter merging with default types

tomsfernandez:
This PR fixes a merging error in:
https://github.com/aml-org/amf/pull/1187/files#diff-06f97a03634e4877d5a3e42a9590918ea7b302853db18558cf3d0b66cd559b47

This is because a default uriParameter is created for myUriParameter of type string. When the resource type is applied the schema with inheritance chain `(myInteger <- integer)` gets merged with the `(string)` resulting in: `(string <- integer)`. This creates an error later on in ShapeCanonization.

When replacing the uriParameter schema type `myInteger` for `integer` there isn't any error because scalars aren't merged.

**Solution**: avoid merging default uriParameter when applying traits/resource types. This enables the user to override undeclared parameters with resource types/traits.

**Note 1**: this might be a breaking change as users that were attempting to override their default uriParameters with resource type and traits didn't have the correct result or they had errors.. APIKit validates with the relaxed validation mode so for scalars the validation result should be the same. Now, uriParameter validation of object uriParameters and arrays will start to kick in.

**Note 2**: several JSON-LD changed because I added the DefaultNode annotations to default uriParameters


nschejtman (approved):

tomsfernandez:
PR Merge fails because of a NumberFormatException in scoverage. All the tests passed.

PR #1239, opened 2021-12-20
Chore && Clean: removed build.js script. Build is entirely with Scala.JS

tomsfernandez:
This PR aims to simplify our use of JS Library imports with Scala JS.

Previously we used a script to expose the Ajv validator in the global space and made the global import from ScalaJS. This is not a good practice because:

- we are polluting the global space of the end user with our dependencies
- when we modify the JS bundle, we are invalidating source maps that are useful for debugging errors

The only downside of correcting this is that we have to declare the Ajv dependency on every module that depends on amf-shapes. Why? Each test module uses the JS bundle of that module and the node packages installed in that module.


amf-cli/js/build-scripts/create-bundle.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/bash
```
AgustinBettati, line 1: js-build.js file can also be removed as it will only be calling ```sbt cliJS/fullOptJS```. If removed we should adjust release process documentation.

tomsfernandez: I left it because I feel it is valuable to have some place where to do extra things besides the fullOptJS. It value although we start with just one command. Besides it standardizes build and publish process.

AgustinBettati (approved):

PR #1275, opened 2022-01-18
APIMF-3603: Fix: Parsing api with json with object as key throws a violation

hghianni:


amf-cli/shared/src/test/scala/amf/validation/AMFModelAssertionTest.scala
```diff
@@ -232,4 +233,16 @@ class AMFModelAssertionTest extends AsyncFunSuite with Matchers {
       inlineShapeAnnotations.lexical().end shouldBe Position(14, 26)
     }
   }
+
+  test("Parsing api with json with object as key should throw a violation") {
```
AgustinBettati: I would suggest using RamlModelUniquePlatformReportTest, you will avoid having to write custom assertions and will save all the details of the generated report.

AgustinBettati (approved):

PR #1291, opened 2022-02-02
APIMF-3584: implemented Vocabulary creation for duplicate terms in schema or properties with multiple characteristics

tomsfernandez:
Related to: https://github.com/aml-org/amf-aml/pull/463

### DIFF
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/SchemaTransformer.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/SchemaTransformer.scala
@@ -1,22 +1,95 @@
 package amf.shapes.internal.spec.jsonschema.semanticjsonschema.transform
 
+import amf.aml.client.scala.model.document.Vocabulary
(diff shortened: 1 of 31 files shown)


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/SchemaTransformer.scala
```diff
@@ -1,22 +1,95 @@
...
 case class TransformationResult(encoded: DomainElement,
                                 declared: Seq[DomainElement],
-                                externals: mutable.Map[String, String])
+                                externals: Map[String, String],
+                                vocab: Option[Vocabulary] = None)
 
-case class SchemaTransformer(shape: AnyShape)(implicit errorHandler: AMFErrorHandler) {
+object SchemaTransformerOptions {
+  val DEFAULT = SchemaTransformerOptions("policy_vocabulary",
+                                         "https://anypoint.mulesoft.com/policy#",
```
looseale: We should put something generic here as default. AMF is open source.
Some uri of aml with "semantic" instead of "policy" (policy is something specific of this use case)

looseale (approved):

PR #1324, opened 2022-03-02
W-10736116: fix transformation for array with any type items

looseale:


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/PropertyShapeTransformer.scala
```diff
@@ -34,11 +32,12 @@ case class PropertyShapeTransformer(property: PropertyShape, ctx: ShapeTransform
     propertyMapping
   }
 
-  private def transformArray(array: ArrayShape) = {
+  private def transformArray(array: ArrayShape): Unit = {
     propertyMapping.withAllowMultiple(true)
     array.items match {
```
tomsfernandez, line 37: Can't we just call the root schema transformation here? A possible blocker is that we don't accept ScalarShapes as roots but can this be a possible case?

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/PropertyShapeTransformer.scala
```diff
@@ -34,11 +32,12 @@ case class PropertyShapeTransformer(property: PropertyShape, ctx: ShapeTransform
     propertyMapping
   }
 
-  private def transformArray(array: ArrayShape) = {
+  private def transformArray(array: ArrayShape): Unit = {
```
tomsfernandez, line 35: What about array of arrays ?

amf-cli/shared/src/test/resources/semantic-jsonschema/dialects/any-array.yaml
```diff
@@ -0,0 +1,20 @@
+#%Dialect 1.0
+nodeMappings:
```
tomsfernandez, line 2: Do we have tests for other types of array types ?

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/PropertyShapeTransformer.scala
looseale: Yes, that is the problem to call the root. Also calling the root transformation will extract the mapping to a declaration, for scalars this makes no sense.

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/semanticjsonschema/transform/PropertyShapeTransformer.scala
looseale: Same as the previous. We should accept it at root level, because we need to extract it to a declaration with the current behavior. I will report another issue to support this, it is not trivial to refactor this.

amf-cli/shared/src/test/resources/semantic-jsonschema/dialects/any-array.yaml
looseale: Same as the previous one. For scalars and objects is tested, no specifical tests but tested in others.

tomsfernandez (approved):

PR #1326, opened 2022-03-03
APIMF-3678: Fix loosing referenced request annotations

hghianni:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/oas/parser/domain/Oas30RequestParser.scala
```diff
@@ -31,7 +31,8 @@ case class Oas30RequestParser(map: YMap, parentId: String, definitionEntry: YMap
       case Left(fullRef) =>
         parseRef(fullRef)
       case Right(_) =>
-        val request = adopt(Request(Annotations.virtual()))
+        val annotations = Annotations.virtual() += BaseVirtualNode(definitionEntry)
```
pope1838, line 34: This is almost a contradiction. If we are adding the ast, then what sense has mark the node as virtual?
Why we need the ast for an object which is not represented by that ast? Maybe we are propagating a bug

tomsfernandez: @llibarona please comment on this

llibarona: The node must be virtual because it contains fields which may or may not be contained in the AST inside this node.
That said, we need the lexical of the base node because it can be referenced in other places and if it is just a virtual there is no way of traking the target reference.
The underlying problem is that a virtual field (request) is used to map a typed entry (requestBody) in oas 3

pope1838 (approved):

PR #1330, opened 2022-03-07
W-10737157: Add warning for format and pattern combination

hghianni:


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/oas/parser/InlineOasTypeParser.scala
```diff
@@ -321,6 +323,21 @@ case class InlineOasTypeParser(entryOrNode: YMapEntryLike,
...
+  private def checkPatternAndFormatCombination(map: YMap, shape: Shape): Unit = {
+    val pattern = map.key("pattern")
+    val format  = map.key("format")
+
+    if (pattern.isDefined && format.isDefined) {
+      val formatName = format.get.value.as[YScalar].text
+      ctx.eh.warning(
+        IgnoredPatternWarning,
+        shape,
+        s"Pattern property is ignored as format '$formatName' already defines a standard pattern",
```
nschejtman: Rephrase: Pattern property MAY BE ignored as format '$formatName' MAY already define a standard pattern

nschejtman: Format property is an arbitrary string. This is only a problem when the format is a known format (date, date-time, email, UUID ,etc) that already defines a pattern

nschejtman (changes requested):

nschejtman (approved):

PR #1347, opened 2022-03-15
W-10663809 - Added in conversion support to extended schemas and if/then/else

looseale:
Depends on https://github.com/aml-org/amf-aml/pull/493


amf-cli/shared/src/test/resources/semantic-jsonschema/dialects/allOf-with-extended-schema.yaml
```diff
@@ -0,0 +1,39 @@
+#%Dialect 1.0
+nodeMappings:
+  Child:
```
tomsfernandez, line 3: Weren't we going to add something to avoid creating this empty mapping ? Don't really care if it is left empty, just asking. Asking because the root schema has properties

tomsfernandez (dismissed):

amf-cli/shared/src/test/resources/semantic-jsonschema/dialects/allOf-with-extended-schema.yaml
looseale: I think is better like this, to have the same behavior in all cases.

tomsfernandez (approved):

PR #1381, opened 2022-04-13
W-10974844 - Changes SemJsonSchema conversion for arrays

looseale:


amf-cli/shared/src/test/resources/semantic-jsonschema/dialects/allOf-complex.yaml
```diff
@@ -25,9 +25,9 @@ nodeMappings:
     mapping:
       a:
         range: any
-        mandatory: false
         enum:
           - condition
+        mandatory: false
```
tomsfernandez, line 30: weird annotation change

looseale: ? It only changes the position. It is because I needed to change the order of the in-memory mapping field creation (now the minCount/mandatory field are populated before or after, depending on it is an array or not)

tomsfernandez (dismissed):

tomsfernandez (approved):

PR #1390, opened 2022-04-22
W-10671730: add graphql directives parsing

arielmirra:
- W-10671730: add GraphQL Locations and Tokens
- chore: refactor graphql helpers
- W-10671730: add GraphQL directive parsing


amf-graphql/shared/src/main/scala/amf/graphql/internal/spec/domain/GraphQLDirectiveDeclarationParser.scala
```diff
@@ -0,0 +1,56 @@
...
+import amf.shapes.client.scala.model.domain.NodeShape
+import org.mulesoft.antlrast.ast.{Node, Terminal}
+
+case class GraphQLDirectiveDeclarationParser(node: Node)(implicit val ctx: GraphQLWebApiContext)
+    extends GraphQLASTParserHelper {
+  var directive: CustomDomainProperty = CustomDomainProperty(toAnnotations(node))
+
+  def parse(parentId: String): CustomDomainProperty = {
+    directive.adopted(parentId)
+    parseName()
```
nschejtman, line 15: You first need to parse the name with is required for ID adoption. You can check that all the directives in is-input-type-directive-arguments.jsonld end up with the same ID and some IDs have nullwrapper and null in them

nschejtman (changes requested):

amf-graphql/shared/src/main/scala/amf/graphql/internal/spec/domain/GraphQLDirectiveDeclarationParser.scala
arielmirra: that's true, I've put the adopt after the `parseName()` method.

nschejtman (approved):

PR #1403, opened 2022-05-10
W-10548016: Fix traits not being applied to some operations of an extension

hghianni:
…ension

### DIFF
--- a/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/stages/ExtendsResolutionStage.scala
+++ b/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/stages/ExtendsResolutionStage.scala
@@ -170,7 +170,7 @@ class ExtendsResolutionStage(profile: ProfileName, val keepEditingInfo: Boolean,
       val resolver = TraitTransformer()
 
       // Iterate operations and resolve extends with inherited traits.
-      val traitList = endpoint.operations.flatMap { operation =>
+      val traitList = endpoint.operations.toList.flatMap { operation =>
         val local = context.add("methodName", operation.method.value())
 
         val branches = ListBuffer[BranchContainer]()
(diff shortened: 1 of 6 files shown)


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/stages/ExtendsResolutionStage.scala
```diff
@@ -170,7 +170,7 @@ class ExtendsResolutionStage(profile: ProfileName, val keepEditingInfo: Boolean,
       val resolver = TraitTransformer()
 
       // Iterate operations and resolve extends with inherited traits.
-      val traitList = endpoint.operations.flatMap { operation =>
+      val traitList = endpoint.operations.toList.flatMap { operation =>
```
nschejtman, line 173: Why is this necessary?

nschejtman: WHAT... this is the only fix? Can you explain?

hghianni: Streams are lazy. In this case when using flatMap for endpoint.operations, which is a Stream, the function is only computed for the first operation. This modifies the endPoint variable by applying the trait to the first operation. As the rest of the values of traitList are never used, the function is never computed for the rest of the operations and the trait is never applied to them. A way to solve is using a not lazy class like Lists.

hghianni: Using traitList.toStream.force also works as this will compute all values of the Stream which translates to function being computed for every operation.

nschejtman: Nice catch!

nschejtman (approved):

PR #1416, opened 2022-05-26
W 10674627: parse GraphQL type extensions

nschejtman:
- Rename GraphQL inheritance parsing -> interface parsing
- W-10674627: parse GraphQL type extensions

When we parse schema extensions we cannot simply add it to the shape `inherits` field. This is because the extended schema might not be present in the API. We will manage this (search the extended schema, add extensions to its `inherits` 
field and compute the minShape) during transformation 

Depends on: https://github.com/aml-org/amf-core/pull/474


antoniogarrote (dismissed):
LGTM!

arielmirra (approved):

PR #1432, opened 2022-06-07
Use JDK 11 in docker

looseale:


gutee:
!gate waiver add Sonar

arielmirra (approved):

PR #1436, opened 2022-06-08
chore: changed sonarqube properties config and coverage generation because of a broken integration with sonarqube scala

tomsfernandez:


gutee:
!gate waiver add Sonar

looseale (approved):

PR #1437, opened 2022-06-08
some fixes

nschejtman:
- W-11267027: fix operation.withResponse(name) interface
- Adopt GovernanceMode on adhoc-cli

### DIFF
--- a/adhoc-cli/src/main/scala/amf/adhoc/cli/Main.scala
+++ b/adhoc-cli/src/main/scala/amf/adhoc/cli/Main.scala
@@ -77,7 +77,7 @@ object Main {
     val renderOptions = if (withLexical) {
       RenderOptions().withPrettyPrint.withSourceMaps.withSourceInformation
     } else {
-      RenderOptions().withPrettyPrint
+      RenderOptions().withPrettyPrint.withGovernanceMode
     }
 
     println {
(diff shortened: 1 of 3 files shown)


adhoc-cli/src/main/scala/amf/adhoc/cli/Main.scala
```diff
@@ -75,9 +75,9 @@ object Main {
     baseUnit = transformation.baseUnit
 
     val renderOptions = if (withLexical) {
-      RenderOptions().withPrettyPrint.withSourceMaps.withSourceInformation
+      RenderOptions().withPrettyPrint.withGovernanceMode.withSourceMaps.withSourceInformation
```
hghianni: The withSourceMaps is unnecessary here

nschejtman: The lexical information is included in the `.withSourceMaps` or `.withSourceInformation` render option?

hghianni: I understand that lexical is included with `.withSourceMap` but the `GovernanceMode` will filter them.

hghianni: So `.withSourceMaps` should be used instead of `withGovernanceMode`

hghianni (dismissed):

hghianni (approved):

PR #1447, opened 2022-06-15
W 11048457

nschejtman:
- W-10671723: Add GraphQL resolution pipelines
- W-11048457: resolve GraphQL type extensions

Notice we are using the EmbeddedJsonLDEmitter for the resolution tests for easier debugging. This emitter has a performance hack that extracts links to declarations, that's why you will see links in some JSON-LDs. The flattened emitter does not have this hack


antoniogarrote (dismissed):
LGTM

looseale (approved):

PR #1449, opened 2022-06-22
W-11332855 & W-11046870

arielmirra:
- W-11046870: add initial validation config for graphql
- W-11046870: add GraphQLUniquePlatformUnitValidationsTest suite
- W-11332855: fix matchError exception in unexpected EOF

### DIFF
--- a/amf-graphql/shared/src/main/scala/amf/graphql/client/scala/GraphQLConfiguration.scala
+++ b/amf-graphql/shared/src/main/scala/amf/graphql/client/scala/GraphQLConfiguration.scala
@@ -3,6 +3,11 @@ package amf.graphql.client.scala
 import amf.antlr.internal.plugins.syntax.{AntlrSyntaxParsePlugin, AntlrSyntaxRenderPlugin}
 import amf.apicontract.client.scala.{AMFConfiguration, APIConfigurationBuilder}
 import amf.apicontract.internal.transformation.{GraphQLCachePipeline, GraphQLEditingPipeline}
+import amf.apicontract.internal.validation.model.ApiEffectiveValidations.GraphQLEffectiveValidations
+import amf.apicontract.internal.validation.model.ApiValidationProfiles.GraphQLValidationProfile
+import amf.apicontract.internal.validation.payload.PayloadValidationPlugin
+import amf.apicontract.internal.validation.shacl.ShaclModelValidationPlugin
+import amf.core.client.common.validation.ProfileNames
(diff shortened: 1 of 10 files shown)


amf-graphql/shared/src/main/scala/amf/graphql/client/scala/GraphQLConfiguration.scala
```diff
@@ -3,19 +3,34 @@ package amf.graphql.client.scala
...
     common()
-      .withPlugins(List(GraphQLParsePlugin, AntlrSyntaxParsePlugin, GraphQLRenderPlugin, AntlrSyntaxRenderPlugin))
+      .withPlugins(
+        List(
+          GraphQLParsePlugin,
+          AntlrSyntaxParsePlugin,
+          GraphQLRenderPlugin,
+          AntlrSyntaxRenderPlugin,
+          ShaclModelValidationPlugin(ProfileNames.GRAPHQL),
+          PayloadValidationPlugin(ProfileNames.GRAPHQL)
```
tomsfernandez: Why is the payload validation plugin necessary ?

arielmirra: it's not for now, will remove it.

amf-graphql/shared/src/main/scala/amf/graphql/internal/spec/parser/syntax/GraphQLASTParserHelper.scala
```diff
@@ -47,11 +47,8 @@ trait GraphQLASTParserHelper extends AntlrASTParserHelper {
     val maybeName = potentialPaths.map { p =>
       path(n, p)
     } collectFirst {
-      case Some(t: Terminal) => t.value
-      case Some(n: Node) =>
-        n.children match {
-          case mutable.Buffer(onlyChild: Terminal) => onlyChild.value
-        }
+      case Some(t: Terminal)                     => t.value
+      case Some(n: Node) if n.children.size == 1 => n.children.head.asInstanceOf[Terminal].value
```
tomsfernandez: is the first value of the nodes children always a terminal ??

arielmirra: in the case of the paths that may have a name, they all do. But that's only the case of a semi-correct API, so I've added an extra check in the match just in case 👌🏽

tomsfernandez (approved):

PR #1451, opened 2022-06-24
W-11178013: complete refactor of the Raml Json Schema Parser, reference parsing in OAS and parsing context methods.

tomsfernandez:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/parser/WebApiContext.scala
```diff
@@ -75,22 +78,28 @@ abstract class WebApiContext(
     with ParseErrorHandler
     with IllegalTypeHandler {
 
-  val syamleh                                                            = new SyamlAMFErrorHandler(wrapped.config.eh)
-  override def handle[T](error: YError, defaultValue: T): T              = syamleh.handle(error, defaultValue)
-  override def handle(location: SourceLocation, e: SyamlException): Unit = syamleh.handle(location, e)
+  private val syamlEh                                                    = new SyamlAMFErrorHandler(wrapped.config.eh)
```
pope1838, line 81: Are you 100% sure that this will not break some adopter? (like ALS)

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/oas/parser/document/Oas3DocumentParser.scala
```diff
@@ -41,7 +50,7 @@ case class Oas3DocumentParser(root: Root)(implicit override val ctx: OasWebApiCo
       parseExamplesDeclaration(map, parent + "/examples")
       parseLinkDeclarations(map, parent + "/links")
       super.parseSecuritySchemeDeclarations(map, parent + "/securitySchemes")
-      super.parseTypeDeclarations(map, parent + "/types", Some(this))
+      super.parseTypeDeclarations(map, Some(this))
```
pope1838, line 53: This could break Visual Designer logic. VS is creating apis using the object oriented model, and trying to replicate our way to generate ids.

tomsfernandez: id adoption inside amf parser code is useless as we apply final id adoption after.

arielmirra (approved):

PR #1456, opened 2022-06-28
W-11235868: update typings

arielmirra:


amf-cli/js/typings/amf-client-js.d.ts
```diff
@@ -1492,6 +1501,8 @@ declare module 'amf-client-js' {
 
     withPayloads(payloads: Array<Payload>): this
 
+    withPayload(): Payload
```
tomsfernandez, line 1504: why add this ?

tomsfernandez (commented):
Did you make sure the typing compiles ?

amf-cli/js/typings/amf-client-js.d.ts
arielmirra: The typings compile, I'm adding that method because it's part of the `Message` class. Also, every class that extends it already has that method (`Request`, `Response`).

tomsfernandez (approved):

PR #1466, opened 2022-07-05
add ADR 11: annotations lifespan interfaces

arielmirra:


adrs/0011-annotations-lifespan.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Context
+Not all annotations are needed in all stages of AMF (parsing, resolution/transformation, validation, rendering/emission).
+
+## Decision
+For this reason, the decision has been made in favor of adding 3 distinct types of annotations, that all others should implement:
+
+- `ResolvableAnnotation`
+    - An annotation that must be resolved
+    - To allow deferred resolution on unordered graph parsing
```
tomsfernandez, line 15: This phrase is pretty but it isn't clear. What is "deferred resolution" ? When is graph parsing "unordered" ?

adrs/0011-annotations-lifespan.md
```diff
@@ -0,0 +1,28 @@
...
+- `SerializableAnnotation`
+    - An annotation that needs to be written in the JSON-LD
```
tomsfernandez: `written` -> `dumped`

adrs/0011-annotations-lifespan.md
```diff
@@ -0,0 +1,28 @@
+# 10. Annotation types
+Date: 2022-07-05
```
tomsfernandez: The date is weird, this decision was taken a long time ago but the ADR was written now. This is shady as doesn't match with the git annotations in the code.

adrs/0011-annotations-lifespan.md
arielmirra: not really sure, extracted it from the javadoc written by Pedro. It's used in the RDFParser.

adrs/0011-annotations-lifespan.md
arielmirra: changed it to the last meaningful commit in the code 👌🏽

arielmirra: changed it to the last meaningful commit in the code 👌🏽

arielmirra: changed it to the last meaningful commit in the code 👌🏽

adrs/0011-annotations-lifespan.md
pope1838: We all agree that graph our graph do not ensure order, but I think that fact has no relation with the definition of a resolvable annotation.
For me, resolvable annotations means that the value of the annotation must be processed into the rdf model while parsing it.

pcolunga: The 'resolvable' name comes from the _theoretical_ resolution/linkage phase of a compiler (usually needed, depending on the complexity of the underlying model, for resolving _references_ within the model).
When parsing from RDF, order cannot be guaranteed for nodes, properties, nor types (as with jsonld but we are taking wrong assumptions there). The idea was to enable an annotation to be resolved after parsing [with some context] (parse -> resolve -> consume).

pcolunga (dismissed):

hghianni (approved):

PR #1471, opened 2022-07-08
W-11047878: Add invalid name validation for GraphQL

hghianni:


amf-graphql/shared/src/main/scala/amf/graphql/internal/spec/parser/syntax/GraphQLASTParserHelper.scala
```diff
@@ -41,21 +41,23 @@ trait GraphQLASTParserHelper extends AntlrASTParserHelper {
       Seq(NAME)
     )
 
-    val maybeName = potentialPaths.map(path(n, _)) collectFirst {
-      case Some(t: Terminal) => t.value
+    val effectivePath = potentialPaths.map(path(n, _)) collectFirst {
```
tomsfernandez: Extract to function

amf-graphql/shared/src/main/scala/amf/graphql/internal/spec/parser/syntax/GraphQLASTParserHelper.scala
```diff
@@ -198,4 +200,23 @@ trait GraphQLASTParserHelper extends AntlrASTParserHelper {
...
+  }
+
+  private def isIntrospectionName(name: String) = name.startsWith("__")
+
+  private def invalidIntrospectionName(name: String, location: Annotations)(implicit
+      ctx: GraphQLWebApiContext
+  ): Unit = {
+    ctx.eh.violation(
+      InvalidIntrospectionName,
+      name,
```
tomsfernandez: The 'target' field in an AMFValidationResult is the element's id, not the value of the evaluated property. It is something like this:
```
  Target: file://amf-cli/shared/src/test/resources/production/inter-spec-refs/oas-raml-securityScheme/api.json#/declares/scheme/oauth20
  Property: http://a.ml/vocabularies/security#type
```

Why can't this go as a custom shacl function instead of a parsing side validation ?

tomsfernandez (changes requested):

tomsfernandez (dismissed):

tomsfernandez (dismissed):

tomsfernandez (approved):

PR #1475, opened 2022-07-11
W 11386887/debt/shape parser context refactor

tomsfernandez:
Depends on: https://github.com/aml-org/amf-core/pull/490

**Why is this change necessary?**

We need to be able to create standalone ShapeParserContext implementations to fix https://github.com/aml-org/amf/pull/1465. Everything related to JsonSchemaDocuments should be in the amf-shapes module but before this there was no way of creating a suitable ParserContext.

The most important commit is the `feat` commit. I left the `ShapeParserContext` interface on purpose as it had another implementation that I didn't want to modify. Sorry its long, I couldn't break it into cohesive atomic parts

- W-11386887 (refactor/test-changed): moved raml external parsers to amf-shapes. Added necessary methods in ShapeParserContext
- W-11386887 (refactor): removed typeParser method from ShapeParserContext
- W-11386887 (refactor): removed closedRamlTypeShape method from contexts. Had to move closed-shape algorithm to amf-shapes
- W-11386887 (feat): created a common implementation of a shape parser context from which all webapi context inherit. I also had to change the Declarations hierarchy as vars and inheritance aren't supported by Scala
- W-11386887 (refactor): removed overlapping 'libraries' variables in WebApiDeclarations


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/WebApiDeclarations.scala
```diff
@@ -33,70 +33,53 @@ import amf.shapes.client.scala.model.domain.Example
...
-          .withEncodes(shape)
-        fragments += (text -> FragmentRef(shape, fragmentRef.location))
-      case _ =>
-        promotedFragments :+= DataTypeFragment().withId(fullRef).withLocation(fullRef).withEncodes(shape)
-        fragments += (text -> FragmentRef(shape, None))
-    }
+    val futureDeclarations: FutureDeclarations
+) extends ShapeDeclarations(alias, errorHandler, futureDeclarations = futureDeclarations) {
+
+  var resourceTypes: Map[String, ResourceType]          = Map()
```
pope1838, line 47: Why do you need to move vars out of class signature?

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/raml/parser/context/RamlSpecAwareContext.scala
```diff
@@ -40,7 +40,7 @@ abstract class RamlSpecVersionFactory(implicit val ctx: RamlWebApiContext) exten
 
   def documentParser: (Root) => RamlDocumentParser
 
-  def typeParser: (YMapEntry, Shape => Unit, Boolean, DefaultType) => RamlTypeParser
+  def typeParser(entry: YMapEntry, adopt: Shape => Unit, isAnnotation: Boolean, default: DefaultType): RamlTypeParser
```
pope1838: Why this change? Keep in mind that this parser build will be different to the rest at other contexts

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
```diff
@@ -0,0 +1,305 @@
...
+import amf.shapes.internal.validation.definitions.ShapeParserSideValidations.{
+  AnnotationSchemaMustBeAny,
+  MissingAnnotationSchema
+}
+import org.mulesoft.common.client.lexical.SourceLocation
+import org.yaml.model._
+
+import scala.collection.mutable
+
+class CommonShapeParseContext(
```
pope1838: This implementation and ShapeParserContext could be one?
I mean, all implementation in here are the same on others web api context?
Also, named just ShapeParserContext maybe?

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/WebApiDeclarations.scala
tomsfernandez: The main reason had to move them is because when adding the intermediate `ShapeDeclarations` class the scala compiler threw an error regarding vars and constructor parameters.

Another reason is that we were duplicating variable declarations for the same things. For example we had two variables for libraries (libs and libraries), two for fragments (frags and fragments) and so on. Moving those variables from the class signature helped me unify things.

Besides, when I saw it there wasn't any class that was using the constructor with all the parameters. The only constructor params being used was the one I left. When classes populated a Declarations instance they did it by copying field by field

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/raml/parser/context/RamlSpecAwareContext.scala
tomsfernandez: This changed because we had the exact same behaviour at the Raml10TypeParser and the Raml08TypeParser. It doesn't make sense to have these at the context level as Raml10 and Raml08 contexts will never be mixed with their parser counterparts. (Raml10 type parser will always want to create a Raml10Type parser and the same with Raml08.)

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
tomsfernandez: Yes, it is an option to unify them but there was another implementation of the ShapeParserContext interface that I didn't want to mess. The context that implementes the interface is in an isolated module that shouldn't bring us troubles (semantic jsonld).

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/raml/parser/context/RamlSpecAwareContext.scala
pope1838: My point is that we were working with functions as return type, and now is returning the parser but just for types. All other methods at web api level keep returning functions to build parsers, not parsers it self

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
pope1838: WebApi works with shapes, my fear is to have an abstract method implemented twice.
For instance, the shapes declaration map (name->shape) should be exactly the same for shapes and web apis, right?
Other case could be documentation or examples, the auxiliary methods for those are the same for shapes and web api.
Basically, web api context should be shape context plus web api things, but all shape related logic shouldn't be different at web api level.

tomsfernandez:
The WebApiContext inherits the CommonShapeParserContext so it should never
override its methods. In the current state, a web api context is a shape
parser context because it inherits from the "common" shape context

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
tomsfernandez: The WebApiContext inherits the CommonShapeParserContext so it should never override its methods. In the current state, a web api context is a shape parser context because it inherits from the "common" shape context

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/raml/parser/context/RamlSpecAwareContext.scala
tomsfernandez: The fact that we return functions to build parsers in this context doesn't make sense. Why ? because we never curry them or pass them as parameters elsewhere, we just call them on the spot.

The way we use them, returning a function that build the parser instead of having a function that returns the parser is just syntax sugar. In fact, is is more clear to have a function to return a parser because of the parameter names. If the developer doesn't which params the "function that returns a function that builds a parser works" then he/she would have to scout the uses to read it.

tomsfernandez: In fact, as functions are first class values in scala this is just syntax sugar as it is the same to define the function or to return a function as in either case someone can still grab the function by value. For example:

```scala

class X {
  def a(a: String): String = { // something}
  
  val function = a // this works
  val result = a("something")
  
  def b: String => String = { // something }
  val result = b("something")
  
  def receivesFunc(func: String => String)
  
  receivesFunc(a) // works
  receivesFunc(b) // works
}
```

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
pope1838: So if web api context inherits from CommonShapeParserContext that inherits from ShapeParserContext, we could have only one parser context and that it

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/raml/parser/context/RamlSpecAwareContext.scala
pope1838: I'm not arguing if one is better from the other. My point is that all this builders returns functions, and this chance is breaking that pattern. That will be more confusing than the function.

tomsfernandez: I changed it.

looseale (approved):

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/CommonShapeParseContext.scala
tomsfernandez: Changed it.

PR #1493, opened 2022-07-15
Add builder for json schema at api configuraiton. Implement registrer…

pope1838:
… parse plugin method at configuration

### DIFF
--- a/amf-api-contract/shared/src/main/scala/amf/apicontract/client/scala/AMFConfiguration.scala
+++ b/amf-api-contract/shared/src/main/scala/amf/apicontract/client/scala/AMFConfiguration.scala
@@ -44,6 +44,7 @@ import amf.shapes.client.scala.config.JsonSchemaConfiguration
 import amf.shapes.client.scala.plugin.JsonSchemaShapePayloadValidationPlugin
 import amf.shapes.internal.annotations.ShapeSerializableAnnotations
 import amf.shapes.internal.entities.ShapeEntities
+import amf.shapes.internal.spec.jsonschema.JsonSchemaParsePlugin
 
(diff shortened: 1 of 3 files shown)


looseale (dismissed):

amf-api-contract/shared/src/main/scala/amf/apicontract/client/scala/AMFConfiguration.scala
```diff
@@ -284,6 +285,9 @@ object APIConfiguration extends APIConfigurationBuilder {
 
   private val name = "API"
 
+  def Schemas(): AMFConfiguration = API()
```
tomsfernandez: This is a weird name because Schemas is not narrow enough. Probably break with the 'spec name' pattern and be able to enrich any configuration?

pope1838: What name do you propose?

tomsfernandez: `JsonSchema` ? jaja Im uneasy about the default being the API() though. I don't think this pattern will create a clear vision about API() being the default though

pope1838: I thought on JsonSchema at the beginning, but I think could close the option to add "other" thinks in a future (another king of schema? not sure).
Maybe will never have another kind of schema based on syaml to support, but add JsonSchema sound  ver restrictive (and not sure if clear that is capable to parse another specs as well)

looseale (approved):

PR #1495, opened 2022-07-19
W-11410812 - Support  in draft 2019 as declarations key along with

looseale:


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/TypeDeclarationParser.scala
```diff
@@ -4,18 +4,50 @@ import amf.aml.internal.parse.common.{DeclarationKey, DeclarationKeyCollector}
...
+    definitionsKeys.flatMap(defKey => parseDeclarationMap(map, defKey, declarationKeysHolder)).toList
+  }
+
+  private def validateMultipleDeclarationKeys(map: YMap, definitionsKeys: Seq[String])(implicit
+      ctx: ShapeParserContext
+  ): Unit = {
+    val foundDefKeys = definitionsKeys.flatMap(map.key(_))
+    if (foundDefKeys.size > 1) { // If there is more than 1 definition key present in the map
+      // TODO add test in the new validation suite
+      ctx.eh.warning(
```
pope1838, line 34: If we are going to parse only one key when both are presents, maybe should be a violation instead of a warning?
I think when some information is lost (not parsed) we should throw violations, because some information is not evaluated (and those ignored ast may be invalid at amf)

looseale: We are parsing all the keys (or at least I suppose I do that)

looseale: `definitionsKeys.flatMap(defKey => parseDeclarationMap(map, defKey, declarationKeysHolder)).toList `
All the declarations in all the keys should be parsing

pope1838 (dismissed):

pope1838 (approved):

PR #1503, opened 2022-07-22
W-11470045 - Change severity of multiple def keys to violation and ad…

looseale:
…ded def key check in ref resolution

### DIFF
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/annotations/Annotations.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/annotations/Annotations.scala
@@ -15,6 +15,16 @@ object ParsedJSONSchema extends AnnotationGraphLoader {
     Some(ParsedJSONSchema(value))
 }
 
+case class DocumentDeclarationKey(key: String) extends EternalSerializedAnnotation {
+  override val name: String  = "document-declaration-key"
+  override val value: String = key
+}
+
+object DocumentDeclarationKey extends AnnotationGraphLoader {
+  override def unparse(value: String, objects: Map[String, AmfElement]): Option[Annotation] =
+    Some(DocumentDeclarationKey(value))
+}
+
 /** Represents parsed RAML Data Type from any type of RAML document. */
 case class ParsedRamlDatatype(rawText: String) extends SerializableAnnotation with PerpetualAnnotation {
   override val name: String  = "parsed-raml-datatype"
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/TypeDeclarationParser.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/TypeDeclarationParser.scala
@@ -1,76 +1,86 @@
 package amf.shapes.internal.spec.common.parser
 
 import amf.aml.internal.parse.common.{DeclarationKey, DeclarationKeyCollector}
+import amf.core.client.scala.model.document.Document
 import amf.core.client.scala.model.domain.AmfScalar
 import amf.core.internal.annotations.DeclaredElement
 import amf.core.internal.metamodel.domain.ShapeModel
+import amf.core.internal.parser.YMapOps
 import amf.core.internal.parser.domain.Annotations
-import amf.core.internal.parser.{YMapOps, _}
 import amf.shapes.client.scala.model.domain.{AnyShape, NodeShape}
+import amf.shapes.internal.annotations.DocumentDeclarationKey
 import amf.shapes.internal.spec.oas.parser.OasTypeParser
 import amf.shapes.internal.validation.definitions.ShapeParserSideValidations.{MultipleDefinitionKey, UnableToParseShape}
-import org.yaml.model.{YMap, YScalar}
+import org.yaml.model.{YMap, YMapEntry, YScalar}
 
 object TypeDeclarationParser {
 
-  def parseTypeDeclarations(map: YMap, definitionsKey: String, declarationKeysHolder: Option[DeclarationKeyCollector])(
-      implicit ctx: ShapeParserContext
-  ): List[AnyShape] = parseTypeDeclarations(map, Seq(definitionsKey), declarationKeysHolder)
+  def parseTypeDeclarations(
+      map: YMap,
+      definitionsKey: String,
+      declarationKeysHolder: Option[DeclarationKeyCollector]
+  )(implicit ctx: ShapeParserContext): List[AnyShape] =
+    parseTypeDeclarations(map, Seq(definitionsKey), declarationKeysHolder, None)
 
   def parseTypeDeclarations(
       map: YMap,
       definitionsKeys: Seq[String],
-      declarationKeysHolder: Option[DeclarationKeyCollector]
+      declarationKeysHolder: Option[DeclarationKeyCollector],
+      document: Option[Document] = None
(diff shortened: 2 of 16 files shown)


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
```diff
@@ -18,29 +18,40 @@ import amf.shapes.internal.validation.definitions.ShapeParserSideValidations.{
...
+  def extract(ref: String, declarationKey: Option[String]): Either[String, String]
 }
 
 trait JsonSchemaRefNameExtraction extends NameExtraction {
   protected def extractShapeName(uriFragment: String, definitionsKey: String): Option[String] = {
     val regex = ("^\\/" + definitionsKey + "\\/(?<shape>[^\\/]+)$").r
     regex.findFirstMatchIn(uriFragment).map(_.group(1))
   }
+
+  protected def validateDeclarationKey(
```
tomsfernandez: Can't this be out of here and inside the `findShapeForReference` function ?

tomsfernandez: My problem with this is that we have to remember to call it on each NameExtraction. We validate the correct ref format ($defs or definitions) before calling name extraction

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
```diff
@@ -18,29 +18,40 @@ import amf.shapes.internal.validation.definitions.ShapeParserSideValidations.{
...
 
 object Draft2019NameExtraction extends JsonSchemaRefNameExtraction {
-  override def extract(ref: String): Either[String, String] = extractShapeName(ref, "\\$defs")
-    .orElse(extractShapeName(ref, "definitions")) match {
-    case Some(some) => Right(some)
-    case None       => Left(s"uriFragment '$ref' must be in the format of '#/definitions/<name>' or '#/$$defs/<name>'")
-  }
+  override def extract(ref: String, declarationKey: Option[String]): Either[String, String] =
+    extractShapeName(ref, "\\$defs").orElse(extractShapeName(ref, "definitions")) match {
+      case Some(some) => validateDeclarationKey(ref, declarationKey, some)
```
tomsfernandez: For some reason you don't have formatting enabled.

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
```diff
@@ -100,10 +111,11 @@ object JsonSchemaLinker {
       ctx: ShapeParserContext
   ): Either[(ValidationSpecification, String), AnyShape] = {
     // TODO: we suppose we have a valid document entry here
-    val version   = JsonSchemaEntry(document.schemaVersion.value()).get
-    val extractor = nameExtractorFor(version)
+    val version        = JsonSchemaEntry(document.schemaVersion.value()).get
+    val declarationKey = document.annotations.find(classOf[DeclarationsKey]).map(_.key)
+    val extractor      = nameExtractorFor(version)
```
tomsfernandez, line 106: I would call declarationKey validation here

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
```diff
@@ -100,10 +111,11 @@ object JsonSchemaLinker {
       ctx: ShapeParserContext
   ): Either[(ValidationSpecification, String), AnyShape] = {
     // TODO: we suppose we have a valid document entry here
-    val version   = JsonSchemaEntry(document.schemaVersion.value()).get
-    val extractor = nameExtractorFor(version)
+    val version        = JsonSchemaEntry(document.schemaVersion.value()).get
+    val declarationKey = document.annotations.find(classOf[DeclarationsKey]).map(_.key)
```
tomsfernandez: Extract to method

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/TypeDeclarationParser.scala
```diff
@@ -1,76 +1,86 @@
...
+      document: Option[Document] = None
   )(implicit ctx: ShapeParserContext): List[AnyShape] = {
-    validateMultipleDeclarationKeys(map, definitionsKeys)
-    definitionsKeys.flatMap(defKey => parseDeclarationMap(map, defKey, declarationKeysHolder)).toList
+    val definitionEntries = definitionsKeys.flatMap(dk => map.key(dk))
+    validateMultipleDeclarationKeys(definitionEntries.size, map, definitionsKeys)
+    definitionEntries.headOption // I will keep only the declarations of the first key (it should be in priority order)
+      .map { de =>
+        // I will annotate the declaration key in the document if present
+        document.foreach(_.annotations += DeclarationsKey(de.key.asScalar.get.text))
```
tomsfernandez: isn't this already done inside the `parseDeclarationMap` method ?

amf-shapes/shared/src/main/scala/amf/shapes/internal/annotations/Annotations.scala
```diff
@@ -15,6 +15,16 @@ object ParsedJSONSchema extends AnnotationGraphLoader {
     Some(ParsedJSONSchema(value))
 }
 
+case class DeclarationsKey(key: String) extends EternalSerializedAnnotation {
```
tomsfernandez: You introduced another annotation so that it can "survive" the json-ld right? I would change the name to avoid clashing with the one used by ALS.

The other way is to remove this one and instead make the one used by ALS an eternal and make the value by the entry key. This way als can still use it and we can dump and parse it from json-ld

tomsfernandez (changes requested):

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
looseale: I did that first, but the problem is that doing that I return always the same error, and in some cases is not specific enough. Putting it here it checks first at least that the link is a correctly formatted declaration link.

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
looseale: Same response as first

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonschema/parser/document/JsonSchemaLinker.scala
looseale: ? I have formatting enabled

tomsfernandez (dismissed):

tomsfernandez (approved):

PR #1506, opened 2022-07-25
W-11047366: validate input types in graphql

arielmirra:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLObjects.scala
```diff
@@ -20,34 +20,55 @@ case class GraphQLObject(node: NodeShape) {
 
   def inherits: Seq[GraphQLObject] = node.inherits.map(_.asInstanceOf[NodeShape]).map(GraphQLObject)
 
+  def isValidInputType: Boolean = {
```
nschejtman: You should check the `isInput` field in Objects

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLArgumentValidator.scala
```diff
@@ -34,12 +68,17 @@ object GraphQLArgumentValidator {
     val properties = node.properties.flatMap { prop =>
       validateDefaultValue(prop.datatype.getOrElse(""), prop.default, prop.property, prop.annotations)
     }
-    properties
+
+    val operations = node.operations.flatMap(_.parameters).flatMap { param =>
```
nschejtman: Did I miss this?
I believe I removed this because I'm already validating these in AbstractParameters rather than starting from an Object. Left the above validation because in directives we are synthesizing an Object, but that Object will never have operations

nschejtman: Suriprised this didn't alter any .report goldens

nschejtman (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLObjects.scala
arielmirra: that's what's done in the `else` block

nschejtman (approved):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/APICustomShaclFunctions.scala
```diff
@@ -118,14 +146,6 @@ object APICustomShaclFunctions extends BaseCustomShaclFunctions {
           }
         }
       },
-      new CustomShaclFunction {
```
nschejtman, line 121: Why delete this?

nschejtman (changes requested):

nschejtman (approved):

PR #1544, opened 2022-08-16
W-11349298: Fix duplicated resource path in OAS

hghianni:


amf-cli/shared/src/test/resources/validations/async20/validations/resources-paths/resources-paths.yaml
```diff
@@ -0,0 +1,50 @@
+asyncapi: '2.0.0'
+info:
+  title: Correlation ID Example
+  version: '1.0.0'
+
+channels:
+  /user:
```
tomsfernandez, line 7: /user{param} is missing

tomsfernandez (approved):

amf-cli/shared/src/test/resources/validations/async20/validations/resources-paths/resources-paths.yaml
tomsfernandez: this is covered by unit tests.

PR #1564, opened 2022-08-26
W-11086896: add @key validations for graphql federation

arielmirra:
- W-11086896: add location to @key parsing in federation
- W-11086896: add @key validations for graphql federation

### DIFF
--- a/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLValidator.scala
+++ b/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLValidator.scala
@@ -1,18 +1,73 @@
 package amf.apicontract.internal.validation.shacl.graphql
 
-import amf.core.client.scala.model.DataType
+import amf.core.client.scala.model.domain.extensions.PropertyShape
 import amf.core.client.scala.model.domain.{DataNode, NamedDomainElement, ScalarNode, Shape}
 import amf.core.internal.metamodel.Field
 import amf.core.internal.metamodel.domain.ScalarNodeModel
-import amf.core.internal.metamodel.domain.extensions.DomainExtensionModel
+import amf.core.internal.metamodel.domain.extensions.{DomainExtensionModel, PropertyShapePathModel}
 import amf.core.internal.parser.domain.Annotations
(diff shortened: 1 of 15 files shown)


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLValidator.scala
```diff
@@ -1,18 +1,60 @@
...
+              key.annotations
+            )
+          else if (n.isInputOnly.value())
+            validationInfo(
+              PropertyShapePathModel.Path,
+              s"Property '${propertyShape.name}' reference by field set can't be from an input type",
+              key.annotations
+            )
+          else {
+            if (path.size == 1) {
```
nschejtman: You need to check that n == path.last.range

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLValidator.scala
```diff
@@ -1,18 +1,60 @@
...
+              validationInfo(
+                PropertyShapePathModel.Path,
+                s"Property '${propertyShape.name}' reference by field set can't be an object type",
+                key.annotations
+              )
+            } else None
+          }
+        case _ => None
```
nschejtman: You need to throw a validation when matching other shapes (eg. UnionShape) except ScalarShape. If I remember correctly all intermediate steps of a PropertyShapePath will be validated as NodeShapes during linking. All except the last propertyShapeRange. That must be a ScalarShape if I'm not mistaken

nschejtman (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLValidator.scala
nschejtman: That condition will not work for this case:
```graphql
type A @key(fields: "b { c }"){
  b: B
}

type B {
  c: C
}

type C {
  str: String
}
```
`path.size == 2` and `c`s range is `C` which is an object

nschejtman (approved):

PR #1583, opened 2022-09-07
W 11690311

hghianni:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/validation/shacl/graphql/GraphQLDirectiveLocationValidator.scala
```diff
@@ -42,6 +43,15 @@ object GraphQLDirectiveLocationValidator {
     } else None
   }
 
+  private def getDirectiveLocations(
+      directiveApplication: DomainExtension
+  ) = {
+    Option(directiveApplication.definedBy) match {
```
tomsfernandez, line 44: could also be:
Option(directiveApplication.definedBy).map(_.domain.map(_.toString)).toList

arielmirra (approved):

PR #1635, opened 2022-10-17
W-11778938: Fix null value location

hghianni:
Related to : https://github.com/aml-org/syaml/pull/68


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/raml/parser/external/json/InlineJsonSchemaParser.scala
```diff
@@ -19,7 +19,7 @@ object InlineJsonSchemaParser extends JsonParsing with ScopedJsonContext with Er
       ctx: ShapeParserContext
   ): AnyShape = {
 
-    val node        = getJsonParserFor(text, valueAST).document().node
+    val node        = getJsonParserFor(text, valueAST).document(keepTokens = true).node
```
tomsfernandez: @pope1838 wasn't this set to false by default for memory & performance reasons a long time ago ?

pope1838 (dismissed):

looseale (approved):

PR #1658, opened 2022-11-15
Issue/1626 Type array is missing in the generated RAML specification

### LINKED ISSUE #1626: Type array is missing in the generated RAML specification
* **I'm submitting a ...**
  - [ ] bug report
  - [ ] feature request
  - [x] question
  - [ ] other

* **What is the current behavior?**

Defining a new endpoint with an array schema, without specifying the shape of the items outputs `type: array`. If the shape of the items is specified, the facet `type` is missing.

I understand there is a type inference when a RAML file is parsed ([Determine Default Types](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#determine-default-types)), but why the `type: array` is missing when the library is used to build a RAML file?

Is there an option to display the type?

The following code:
```javascript
import amf from 'amf-client-js'

async function main () {
  const api = new amf.WebApi()

  api
    .withEndPoint('/scalar_array')
    .withOperation('post')
    .withRequest()
    .withPayload()
    .withMediaType('application/json')
    .withArraySchema('schema')
    .withItems(new amf.ScalarShape().withName('items').withDataType('http://www.w3.org/2001/XMLSchema#string'))

  const document = new amf.Document().withEncodes(api)

  const client = amf.RAMLConfiguration.RAML10().baseUnitClient()

  const generated = client.render(document)

  console.log(generated)
}

main()
```

produces this output:

```raml
#%RAML 1.0
/scalar_array:
  post:
    body:
      application/json:
        items:
          type: string
```

* **What is the expected behavior?**

`type: array` is expected to be present in the output:

```raml
#%RAML 1.0
/scalar_array:
  post:
    body:
      application/json:
        type: array
        items:
          type: string
```

* **What is the motivation/use case for changing the behavior? (for feature requests)**

Documentation generators like raml2html use the `type` to correctly render the type declarations.

* **Please tell us about your environment:**
  
  - AMF Version: 5.1.0-0
  - AMF Distribution: JS 
  - Operating System: linux x64

f-schnabel:
Fixes #1626
This PR contains both the fix for arrays, but also objects.
If only arrays should be fixed, please comment below.

### DIFF
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/raml/emitter/RamlArrayShapeEmitter.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/raml/emitter/RamlArrayShapeEmitter.scala
@@ -23,7 +23,8 @@ case class RamlArrayShapeEmitter(array: ArrayShape, ordering: SpecOrdering, refe
 
     fs.entry(ArrayShapeModel.Items)
       .foreach(_ => {
-        typeEmitted = true
+        if (spec.options.implicitRamlTypes)
+          typeEmitted = true
         result += RamlItemsShapeEmitter(array, ordering, references)
       })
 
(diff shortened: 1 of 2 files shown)


nschejtman:
@Shadow-Devil the snapshot with the new render options has already published. You can adopt it now on this PR

f-schnabel:
@nschejtman I adopted the PR and force-pushed it. Locally it still doesn't work for me, but that's probably because Gradle cached the old version of amf-core without the new `Renderoption.implicitRamlTypes`.

nschejtman:
> @nschejtman I adopted the PR and force-pushed it. Locally it still doesn't work for me, but that's probably because Gradle cached the old version of amf-core without the new `Renderoption.implicitRamlTypes`.

Funny enough there was a bug, the `implicitRamlTypes` could never be set to false. Fixed that already an opened this PR https://github.com/aml-org/amf/pull/1668 to replace this PR

f-schnabel:
Oh, I'm sorry to introduce the bug.
Hope everything works now!

PR #1663, opened 2022-11-16
W-12005689: add required field in PayloadModel

damianpedra:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/parser/ParameterParsers.scala
```diff
@@ -501,6 +501,7 @@ case class Oas2ParameterParser(
       entry => {
         val req: Boolean = entry.value.as[Boolean]
         payload.annotations += RequiredParamPayload(req, entry.range)
```
nschejtman, line 503: We were in fact parsing required as an annotation. Maybe we could remove this on our next major version. You can change the OAS emiiter to use the new field instead of the annotation

nschejtman (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/parser/ParameterParsers.scala
nschejtman: Also, please update failing JSON-LDs

nschejtman (dismissed):

nschejtman (approved):

PR #1667, opened 2022-11-23
W-11337671

arielmirra:
- W-11337671: add inferred annotation to union examples field
- W-11337671: avoid parsing empty examples
- W-11337671: stop ignoring null as an example value
- W-11337671: move NilUnion examples to union level
- (refactor): move BaseUnitUtils to its own class

### DIFF
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleDataParser.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleDataParser.scala
@@ -38,8 +38,9 @@ case class ExampleDataParser(entryLike: YMapEntryLike, example: Example, options
     }
 
     node.toOption[YScalar] match {
-      case Some(_) if node.tagType == YType.Null =>
-        example.setWithoutId(ExampleModel.Raw, AmfScalar("null"), Annotations.synthesized())
+      case Some(value) if node.tagType == YType.Null =>
+        if (isNullLiteral(value))
+          example.setWithoutId(ExampleModel.Raw, AmfScalar("null"), Annotations.synthesized())
       case Some(scalar) =>
         example.setWithoutId(ExampleModel.Raw, AmfScalar(scalar.text), Annotations.synthesized())
       case _ =>
@@ -55,6 +56,8 @@ case class ExampleDataParser(entryLike: YMapEntryLike, example: Example, options
 
     example
   }
+
+  private def isNullLiteral(value: YScalar) = value.text.nonEmpty
 }
 
 case class ExamplesDataParser(seq: YSequence, options: ExampleOptions, parentId: String)(implicit
--- a/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleParsers.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleParsers.scala
@@ -188,14 +188,26 @@ case class RamlSingleExampleParser(
           node.tagType match {
             case YType.Map =>
               Option(RamlSingleExampleValueParser(entry, newProducer, options).parse())
+            case YType.Null => checkEmptyExample(newProducer, entry, node)
             case _ => // example can be any type or scalar value, like string int datetime etc. We will handle all like strings in this stage
-              Option(
-                ExampleDataParser(YMapEntryLike(node), newProducer().add(Annotations(entry.value)), options)
-                  .parse()
-              )
+              parseExample(newProducer, entry, node)
           }
       }
     }
+  }
+
+  private def checkEmptyExample(newProducer: () => Example, entry: YMapEntry, node: YNode): Option[Example] = {
+    node.value match {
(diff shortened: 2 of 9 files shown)


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/oas/parser/InlineOasTypeParser.scala
```diff
@@ -99,7 +100,15 @@ case class InlineOasTypeParser(
   protected def isOas: Boolean  = version.isInstanceOf[OASSchemaVersion]
   protected def isOas3: Boolean = version.isInstanceOf[OAS30SchemaVersion]
 
-  def checkNilUnion(parsed: AnyShape): AnyShape = {
+  protected def moveExamplesToUnion(parsed: AnyShape, union: UnionShape): Unit = {
```
tomsfernandez, line 103: Shouldn't this be in transformation ??

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleDataParser.scala
```diff
@@ -38,8 +38,9 @@ case class ExampleDataParser(entryLike: YMapEntryLike, example: Example, options
     }
 
     node.toOption[YScalar] match {
-      case Some(_) if node.tagType == YType.Null =>
-        example.setWithoutId(ExampleModel.Raw, AmfScalar("null"), Annotations.synthesized())
+      case Some(nullValue) if node.tagType == YType.Null =>
+        if (nullValue.text.nonEmpty)
```
tomsfernandez: could be something like: "isNullLiteral"

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleParsers.scala
```diff
@@ -188,15 +188,30 @@ case class RamlSingleExampleParser(
...
+              parseExample(newProducer, entry, node)
           }
       }
     }
   }
+
+  private def checkEmptyExample(newProducer: () => Example, entry: YMapEntry, node: YNode): Option[Example] = {
+    node.value match {
+      case value: YScalar =>
+        if (value.text.nonEmpty) {
```
tomsfernandez: "If" could be inlined in match-case

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/common/parser/ExampleParsers.scala
```diff
@@ -325,7 +340,8 @@ case class NodeDataNodeParser(
           }
         )
       case Some(scalar) if XMLSchema.unapply(scalar.text).isDefined => None
-      case _                                                        => Some(node) // return default node for xml too.
+//      case Some(value) if value.text.isEmpty => None
```
tomsfernandez: If not used, erase commented line

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/oas/parser/InlineOasTypeParser.scala
```diff
@@ -101,9 +101,9 @@ case class InlineOasTypeParser(
   protected def isOas3: Boolean = version.isInstanceOf[OAS30SchemaVersion]
 
   protected def moveExamplesToUnion(parsed: AnyShape, union: UnionShape): Unit = {
-    val AmfArray(values, fieldAnnotations) = parsed.fields.get(ExamplesField.Examples)
+    val AmfArray(values, annotations) = parsed.fields.get(ExamplesField.Examples)
     if (values.nonEmpty) {
-      union.setWithoutId(ExamplesField.Examples, AmfArray(values), fieldAnnotations)
+      union.setWithoutId(ExamplesField.Examples, AmfArray(values, annotations), annotations)
```
tomsfernandez: shouldn't this be syntethized, virtual or something like that?

tomsfernandez (changes requested):

tomsfernandez:
Avoid changing the test on each commit

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/oas/parser/InlineOasTypeParser.scala
arielmirra: it should actually be inferred, because it's content it's explicit in the document, just not exactly in an union

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/oas/parser/InlineOasTypeParser.scala
arielmirra: not really, if the union is created in parsing it should be created correctly, not fixed in a later stage

tomsfernandez: ahh because they are nullable fields.ok

tomsfernandez (approved):

PR #1671, opened 2022-12-05
W-11918764

hghianni:


amf-cli/shared/src/test/resources/org/raml/parser/examples/builtin/date-validation/output.txt.jvm
```diff
@@ -14,15 +14,15 @@ Level: Violation
   Location: file://amf-cli/shared/src/test/resources/org/raml/parser/examples/builtin/date-validation/input.raml
 
 - Constraint: http://a.ml/vocabularies/amf/validation#example-validation-error
-  Message: [25:61:61] is not a valid time. Expected HH:mm:ss
+  Message: [25:61:61] is not a valid time. Expected HH:mm:ss[.ff...]
```
tomsfernandez, line 17: Unrelated: the error message isn't the best as it doesn't tell me what is wrong with my timestamp

tomsfernandez (approved):

looseale (approved):

amf-cli/shared/src/test/resources/org/raml/parser/examples/builtin/date-validation/output.txt.jvm
looseale: The regex validators return a boolean value, not a specific error telling why it is invalid. We will not add extra code to get that.

hghianni: I agree with @looseale. Message is not clear as it can be but changing it means doing a manual validation to determine the error.

PR #1695, opened 2023-01-03
Add setters for property and values at jsonldobject, cloning the inst…

pope1838:
…ance and the metadata

### DIFF
--- a/amf-shapes/shared/src/main/scala/amf/shapes/client/platform/model/domain/jsonldinstance/JsonLDObject.scala
+++ b/amf-shapes/shared/src/main/scala/amf/shapes/client/platform/model/domain/jsonldinstance/JsonLDObject.scala
@@ -4,13 +4,38 @@ import amf.core.client.platform.model.domain.DomainElement
 import amf.shapes.client.scala.model.domain.jsonldinstance.{JsonLDObject => InternalJsonLDObject}
 import amf.shapes.internal.domain.metamodel.jsonldschema.JsonLDEntityModel
 import amf.shapes.internal.spec.jsonldschema.parser.JsonPath
+import amf.shapes.internal.convert.ShapeClientConverters._
 
 import scala.scalajs.js.annotation.{JSExportAll, JSExportTopLevel}
 
 @JSExportAll
 class JsonLDObject(override private[amf] val _internal: InternalJsonLDObject) extends DomainElement with JsonLDElement {
   @JSExportTopLevel("JsonLDObject")
-  def this() = this(InternalJsonLDObject.empty(new JsonLDEntityModel(Nil, Nil, JsonPath.empty), JsonPath.empty))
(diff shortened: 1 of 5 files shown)


amf-shapes/shared/src/main/scala/amf/shapes/client/platform/model/domain/jsonldinstance/JsonLDObject.scala
```diff
@@ -4,13 +4,38 @@ import amf.core.client.platform.model.domain.DomainElement
...
+
+  def withProperty(property: String, value: Integer): JsonLDObject = _internal.withProperty(property, value)
+
+  def withProperty(property: String, value: Float): JsonLDObject = _internal.withProperty(property, value)
+
+  def withProperty(property: String, value: Boolean): JsonLDObject = _internal.withProperty(property, value)
+
+  def withProperty(property: String, value: JsonLDObject): JsonLDObject = _internal.withProperty(property, value)
+
+  def withStringPropertyCollection(property: String, values: Seq[String]): JsonLDObject =
```
tomsfernandez, line 27: can't we use method overloading instead of having a different name ?

amf-shapes/shared/src/main/scala/amf/shapes/client/scala/model/domain/jsonldinstance/JsonLDObject.scala
```diff
@@ -22,4 +24,68 @@ class JsonLDObject(
...
+  private def updateModel(field: Field) = model.copy(fields = model.fields.filterNot(_ == field) :+ field)
+  private def copyWithProperty(field: Field, element: JsonLDElement) =
+    copy(fields = fields.copy(), model = updateModel(field)).set(field, element)
+
+  def withProperty(property: String, value: String): JsonLDObject =
+    copyWithProperty(property.toStrField, buildString(value))
+
+  def withProperty(property: String, value: Int): JsonLDObject =
+    copyWithProperty(property.toIntField, buildInteger(value))
+
```
tomsfernandez, line 67: same as last comment

tomsfernandez (approved):

amf-shapes/shared/src/main/scala/amf/shapes/client/platform/model/domain/jsonldinstance/JsonLDObject.scala
pope1838: That was my original approach, but compilation failed :S

amf-shapes/shared/src/main/scala/amf/shapes/client/scala/model/domain/jsonldinstance/JsonLDObject.scala
pope1838: Didn't understood, sorry. In this case the method name is the same for all scalar, the type of the "value" param is the only difference, but that's to avoid do an internal match from "any" (and be forced to handle the unrecognized types)

PR #1752, opened 2023-03-20
W-11460908 [fix regression]: parsing & validating discriminator value as a DataNode in the model

nschejtman:


amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/raml/parser/Raml10TypeParser.scala
```diff
@@ -1605,7 +1605,13 @@ sealed abstract class RamlTypeParser(
 
       map.key("discriminator", (NodeShapeModel.Discriminator in shape).allowingAnnotations)
       map.key("discriminatorValue", (NodeShapeModel.DiscriminatorValue in shape).allowingAnnotations)
-
+      map.key(
+        "discriminatorValue",
```
pope1838, line 1608: You are looking twice for the entry "discriminatorValue". You should do just one search and expand the parsing logic at the entry function.

nschejtman: Look ups are constant time (map.entries is an IndexedSeq). Do you believe it's worth adding the extra lines of code rather than reusing the helper above?

pope1838: The time can be constant, but is duplicated. Other option could be to search the entry, stored at an Optional variable, and the do the map twice, but with the entry already found

pope1838 (approved):

PR #1800, opened 2023-05-09
W-11460908: remove full JSON path from property terms in JSON-LD objects

nschejtman:


pope1838 (dismissed):

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonldschema/parser/JsonLDObjectElementParser.scala
```diff
@@ -39,11 +39,19 @@ case class JsonLDObjectElementParser(
   override def findClassTerm(ctx: SemanticContext): Seq[String] = {
     val terms = super.findClassTerm(ctx).map(ctx.expand)
     if (terms.isEmpty) {
-      val fragment = if (path.toString.isEmpty) key else path.toString
+      val fragment = lastNameFromPath() match {
```
tomsfernandez, line 42: this whole if-clause could be refactored into a method so we isolate @id computation logic

amf-shapes/shared/src/main/scala/amf/shapes/internal/spec/jsonldschema/parser/JsonLDPropertyParser.scala
```diff
@@ -116,10 +116,16 @@ case class JsonLDPropertyParser(
...
-      .getOrElse(semanticContext.computeBase(path.toString))
+    propertyName(path) match {
+      case Some(name) =>
+        semanticContext
+          .getTermFor(name)
+          .map(semanticContext.expand)
+          .getOrElse(semanticContext.computeBase(name))
+
+      // should be unreachable
+      case _ => throw new IllegalStateException(s"Property ${path.toString} does not have name")
```
tomsfernandez, line 127: [Out of Scope] - This is shady, we should have a safe-compiler way of doing this check (I don't know how at the moment)

tomsfernandez (approved):

PR #1842, opened 2023-08-02
W-13813016: avoid re adopt web api at examplesresolution stage

pope1838:


amf-cli/shared/src/test/resources/validations/jsonld-compact-uris/parsed-result.flattened.jsonld
```diff
@@ -120,6 +120,281 @@
         "@id": "#1"
       }
     },
+    {
```
tomsfernandez, line 123: Why do we have an entire new shape here ?

amf-cli/shared/src/test/resources/resolution/oas30-discriminator-json-ld/api.target.flattened.jsonld
```diff
@@ -8,7 +8,7 @@
       "doc:transformed": true
     },
     {
-      "@id": "#2",
```
tomsfernandez, line 11: This is weird, where did the "#2" id go? In an incremental id shortening we "shouldn't" be skipping ids.

pope1838: The encodes is one of the first fields that we set when parsing, so is one of the last in the fields map.
The change was stop doing a withEncodes at resolutions, that moves the encodes field to the top of the map and making it the first to be iterated when generating incremental id. Now preserve the original position, so that can explain why the id now is 17(much bigger).
Didn't check it in runtime, but makes complete sense.

tomsfernandez (approved):

PR #1893, opened 2023-10-27
W-13992415: fix oas to oas nullable parameter emission

arielmirra:


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas3/CleanNullSecurity.scala
```diff
@@ -1,50 +0,0 @@
-package amf.apicontract.internal.transformation.compatibility.oas3
```
tomsfernandez, line 1: What is this change?

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
```diff
@@ -22,11 +22,25 @@ class CompatibilityCycleTest extends AsyncFunSuite with FileAssertionTest {
...
+      "raml10/union-nil-plus-element.raml",
+      "oas30/union-nil-plus-element.yaml",
+      Raml10YamlHint,
+      Oas30YamlHint
+    )
   }
 
   test("RAML with union of nil + multiple types to OAS 3.0") {
-    compatibility("raml10/union-nil-plus-elements.raml", "oas30/union-nil-plus-elements.yaml", Raml10YamlHint, Oas30YamlHint)
+    compatibility(
```
tomsfernandez, line 34: Avoid adding formatting changes in the commit please

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
```diff
@@ -22,11 +22,25 @@ class CompatibilityCycleTest extends AsyncFunSuite with FileAssertionTest {
...
+      "raml10/union-nil-plus-elements.raml",
+      "oas30/union-nil-plus-elements.yaml",
+      Raml10YamlHint,
+      Oas30YamlHint
+    )
+  }
+
+  test("OAS 3.0 with nullable parameter to JSON") {
```
tomsfernandez, line 42: Why would we run the compatibility pipeline for this? Compatibility pipeline is for Spec conversion, not syntax conversion.

tomsfernandez (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas3/CleanNullSecurity.scala
arielmirra: The transformation step was duplicated, once in oas2 and once in oas3, they both did the same, so I deleted the duplicated one.

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
arielmirra: sorry I don't like unformatted code, but didn't expected it would make the code hard to read for you

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
arielmirra: Because this comes from a CRI in which a client uses APID to transfrom from yaml to json, going through the compatibility pipeline, thus encountering this error.

Other discussion if this should be the correct pipeline, but tried other pipelines and got the same result if that helps.

tomsfernandez: Hmm this isn't convincing, as the use case per se is wrong. I don't think it is wise to have tests for unsupported use cases. If the customer wants to go from YAML to JSON then they shouldn't use compatibility pipeline.

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
tomsfernandez: please add it in another commit.

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas3/CleanNullSecurity.scala
tomsfernandez: This should go in another commit too, it is an unrelated change to the functionality you are fixing.

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
tomsfernandez: The problem with this is that if another member of the team makes a change that breaks this test, they will not know that this is an unsupported/incorrect use case and therefore will try to maintain it, ending up in having to maintain something that we don't want to.

tomsfernandez: we discussed it internally, it would be better if we add this as an "editing" case instead of a "compatibility" case.

tomsfernandez: after discussing it internally, its ok to me.

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas3/CleanNullSecurity.scala
arielmirra: done

amf-cli/shared/src/test/scala/amf/emit/CompatibilityCycleTest.scala
arielmirra: done

tomsfernandez (approved):

PR #1950, opened 2024-03-15
W-12689966: add pulsar binding

arielmirra:
- **W-12689966: add pulsar binding model and interfaces**
- **W-12689966: add Pulsar binding parsing**
- **W-12689966: validate pulsar channel binding required fields**
- **W-12689966: add pulsar binding emission**


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/parser/CustomSyntax.scala
```diff
@@ -9,6 +9,12 @@ trait CustomSyntax {
   def contains(shape: String): Boolean = nodes.contains(shape)
 }
 
+object CustomSyntax {
```
looseale: This seems to be not used

looseale (approved):

PR #2055, opened 2024-10-10
W-16888404: change parameter description parsing and bump model

arielmirra:
- **update amf snapshot in js package.json**
- **W-16888404: add avro tck test with payload in external ref**
- **W-16888404: parse server variable descriptions into parameter instead of schema**
- **W-16888404: bump amf model to 3.10.0 for avro model changes**

### DIFF
--- a/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/emitter/ServerEmitters.scala
+++ b/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/emitter/ServerEmitters.scala
@@ -2,14 +2,13 @@ package amf.apicontract.internal.spec.common.emitter
 
 import amf.apicontract.client.scala.model.domain.api.Api
 import amf.apicontract.client.scala.model.domain.{EndPoint, Operation, Parameter, Server}
-import amf.apicontract.internal.metamodel.domain.ServerModel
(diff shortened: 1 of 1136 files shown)


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/emitter/ServerEmitters.scala
```diff
@@ -280,7 +280,7 @@ case class OasServerVariableEmitter(variable: Parameter, ordering: SpecOrdering)
     val shape  = variable.schema
     val fs     = shape.fields
 
-    fs.entry(ShapeModel.Description).map(f => result += ValueEmitter("description", f))
+    variable.fields.entry(ShapeModel.Description).map(f => result += ValueEmitter("description", f))
```
looseale: This should be ParameterModel.Description

looseale (changes requested):

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/common/emitter/ServerEmitters.scala
arielmirra: Done

looseale (approved):

PR #2062, opened 2024-10-28
W-10548506: add OAS 3.1 basic support

arielmirra:
- **W-10548506: add OAS 3.1 config and basic support**
- **W-10548506: add OAS 3.1 parsers**
- **W-10548506: add OAS 3.1 emitters**
- **W-10548506: add Oas31CycleTest for OAS 3.1**


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/spec/oas/parser/document/Oas3DocumentParser.scala
```diff
@@ -38,11 +42,6 @@ case class Oas3DocumentParser(root: Root)(implicit override val ctx: OasWebApiCo
     YamlTagValidator.validate(root)
     val api = super.parseWebApi(map)
     AnnotationParser(api, map).parseOrphanNode("components")
-
-    map.key("consumes".asOasExtension, WebApiModel.Accepts in api)
```
looseale, line 42: Do not delete this. This is to support cycles between specs with amf extensions

looseale: It is already present in the super class OAS2DocParser

looseale (approved):

PR #2100, opened 2025-03-06
W-17759271: fix raml file upload param to oas

arielmirra:
- **chore: improve FixConsumesAndProduces.scala**
- **W-17759271: fix raml file upload param to oas 2.0**
- **W-17759271: fix raml file upload param to oas 3.0**

### DIFF
--- a/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas/FixFilePayloads.scala
+++ b/amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas/FixFilePayloads.scala
@@ -1,0 +1,61 @@
+package amf.apicontract.internal.transformation.compatibility.oas
+
+import amf.apicontract.client.scala.model.domain.Payload
+import amf.apicontract.client.scala.model.domain.api.WebApi
+import amf.apicontract.internal.annotations.FormBodyParameter
+import amf.core.client.scala.AMFGraphConfiguration
+import amf.core.client.scala.errorhandling.AMFErrorHandler
+import amf.core.client.scala.model.document.{BaseUnit, Document}
+import amf.core.client.scala.transform.TransformationStep
+import amf.shapes.client.scala.model.domain.{FileShape, NodeShape}
+
+/** To represent a method with file upload:
+  *
+  * in RAML: \- You define a body with multipart/form-data and a property of type: file \- You can specify acceptable
+  * file types directly with fileTypes: ['application/xml']
+  *
+  * In OAS 2.0 you need a parameter of type: file and in: formData with a consumes: - multipart/form-data, to correctly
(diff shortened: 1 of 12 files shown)


amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas/FixFilePayloads.scala
```diff
@@ -0,0 +1,86 @@
...
+          case p if hasFileUpload(p) => checkFormData(p)
+          case _                     => // ignore
+        }
+      }
+    }
+  }
+
+  private def hasFileUpload(payload: Payload): Boolean = {
+    payload.schema match {
+      case obj: NodeShape => obj.properties.exists(prop => prop.range.isInstanceOf[FileShape])
```
looseale: Why check only for a NodeShape with a property file type? In the issue particular case the type was like this, but this is not always like that.
In RAML spec you have examples with file type at root level, and nothing limitates where it could be defined: https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#applying-libraries
The solution should be general for all possible cases

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas/FixFilePayloads.scala
```diff
@@ -0,0 +1,86 @@
...
+    if (payload.mediaType.value().equals(formData)) {
+      payload.withInferredName("formData")
+      payload.add(FormBodyParameter())
+    }
+  }
+}
+
+case class FixFilePayloadsOas3() extends FixFilePayloads() {
+  override def checkFormData(payload: Payload): Unit = {
+    val formData = "multipart/form-data"
```
looseale: Why limitate this only to multipart/form-data? There are other examples in the spec using file type with other mediatypes: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#considerations-for-file-uploads

amf-api-contract/shared/src/main/scala/amf/apicontract/internal/transformation/compatibility/oas/FixFilePayloads.scala
```diff
@@ -0,0 +1,86 @@
...
+      case _              => false
+    }
+  }
+
+  def checkFormData(payload: Payload): Unit
+}
+
+case class FixFilePayloadsOas2() extends FixFilePayloads() {
+  override def checkFormData(payload: Payload): Unit = {
+    val formData = "multipart/form-data"
```
looseale: Duplicated line, could be extracted to parent

looseale (changes requested):

looseale (approved):

